PR #14, opened 2023-03-15
virtual ptdf

### LINKED ISSUE #2: PTDF and LODF matrices ignore available flag
Currently, the implementation of PTDF and LODF ignore the available flag on the branches.

### LINKED ISSUE #3: Build LODF given PTDF
Building the PTDF can be very time and memory consuming. 
The current LODF function involves building the PTDF, but if we already have calculated the PTDF, this seems like a lot of unnecessary computational effort. It would be a useful feature to have an LODF method that receives a `PTDF` type and uses it.

### LINKED ISSUE #6: Implement Virtual PTDF object

jd-lara:
closes #6 
closes #2 
closes #3


src/incedence_matrix.jl
rodrigomha, line 1: the file name has a typo, should be `incidence_matrix.jl`

rodrigomha (changes requested):
Besides the naming change I think it looks okay

PR #15, opened 2023-03-17
Jd/adjacency matrix

jd-lara:
Add Adjacency Matrix as a first class object


src/adjancency_matrix.jl
```diff
@@ -0,0 +1,166 @@
...
+    end
+    return connected
+end
+
+function dfs_connectivity(M::SparseArrays.SparseMatrixCSC,
+    bus_lookup::Dict{Int, Int})
+    @info "Validating connectivity with depth first search (network traversal)"
+    cc = find_connected_components(M, bus_lookup)
+    if length(cc) != 1
+        @warn "Network has at least $(length(cc)) connected components with $(length.(cc)) nodes"
```
daniel-thom, line 138: Does this error message make sense? Is the return of find_connected_components not guaranteed?

src/adjancency_matrix.jl
```diff
@@ -0,0 +1,166 @@
...
+        #@assert n < node_count - 1
+    end
+    return I
+end
+
+function goderya_connectivity(M::SparseArrays.SparseMatrixCSC{Int8, Int},
+    bus_lookup::Dict{Int, Int})
+    @info "Validating connectivity with Goderya algorithm"
+    node_count = length(bus_lookup)
+    node_count > 15_000 &&
```
daniel-thom, line 113: Can this be a constant defined somwhere with a comment?

src/adjancency_matrix.jl
```diff
@@ -0,0 +1,166 @@
...
+        @warn "The Goderya algorithm is memory intensive on large networks and may not scale well, try `connectivity_method = dfs_connectivity"
+
+    I = _goderya(M)
+
+    connections = Dict([i => count(x -> x == i, I) for i in Set(I)])
```
daniel-thom, line 118: Why are the square brackets here? Instead of
```
connections = Dict(i => count(x -> x == i, I) for i in Set(I))
```

src/adjancency_matrix.jl
```diff
@@ -0,0 +1,166 @@
...
+            break
+        elseif length(I) < max_I
+            temp = T_ * T
+            I_temp, _, _ = SparseArrays.findnz(temp)
+            if all(I_temp == I)
+                @warn "The system contains islands" maxlog = 1
+            end
+            T_ = temp
+        else
+            @assert false
```
daniel-thom, line 102: Call `error` in case assert is ever compiled out.

src/adjancency_matrix.jl
jd-lara: Probably a legacy mistake.

src/adjancency_matrix.jl
jd-lara: This function has some issues that require a re-write from scratch. That needs a separate PR that relies on this one.

daniel-thom (approved):

PR #25, opened 2023-03-27
Doc string upadate

alefcastelli:
Added the documentation related to each function in the Package.

y_bus still to finish.


src/virtual_ptdf_calculations.jl
```diff
@@ -231,7 +331,37 @@ function _getindex(
...
+    row_, column_ = to_index(vptdf, row, column)
+    return _getindex(vptdf, row_, column_)
+end
+
+# Define for ambiguity resolution
+function Base.getindex(vptdf::VirtualPTDF, row::Integer, column::Integer)
+    return _getindex(vptdf, row, column)
+end
+
+"""
```
jd-lara, line 357: We don't need this for unsupported methods

src/virtual_ptdf_calculations.jl
```diff
@@ -174,50 +284,40 @@ function Base.isempty(vptdf::VirtualPTDF)
...
-    return _getindex(vptdf, row_, column_)
-end
+"""
+Funciton for internal use only.
 
-# Define for ambiguity resolution
-function Base.getindex(vptdf::VirtualPTDF, row::Integer, column::Integer)
-    return _getindex(vptdf, row, column)
-end
+Gets the value of the element of the PTDF matrix given the row and column indices
```
jd-lara, line 304: This is an internal function that doesn't require a docstring. Just documenting `Base.getindex` is enough

src/virtual_ptdf_calculations.jl
```diff
@@ -174,50 +284,40 @@ function Base.isempty(vptdf::VirtualPTDF)
     !isempty(vptdf.temp_data) && return false
     return true
 end
-# Size related overload will work on the BA matrix
+
+"""
+Gives the size of the whole PTDF matrix, not the number of rows stored.
```
jd-lara, line 289: Check the wording for `Base.size` it reads 

```suggestion
Return a tuple containing the dimensions of the VirtualPTDF, not the number of rows stored.
```

src/virtual_ptdf_calculations.jl
```diff
@@ -148,21 +259,20 @@ function VirtualPTDF(
...
     branches = get_ac_branches(sys)
-    nodes = get_buses(sys)
+    buses = get_buses(sys)
 
-    return VirtualPTDF(branches, nodes; kwargs...)
+    return VirtualPTDF(branches, buses; kwargs...)
 end
 
-# Overload Base functions
+"""
```
jd-lara, line 273: There is a typo and also, check how the default is written 

```suggestion
"""
Determine whether the VirtualPTDF is empty (has no elements).
```

jd-lara (changes requested):
- [x] Only add docstrings for the exported method. 
- [ ] Check the format for the docstrings https://docs.julialang.org/en/v1/manual/documentation/. Note that we use DocstringExtensions.jl so there is no need to write the function signature. 
- [x] There are language typos

PR #37, opened 2023-04-20
Serialize PTDF to HDF5 files

daniel-thom:


daniel-thom:
@jd-lara I don't understand why the performance comparison was canceled. Does that normally work?

jd-lara:
@daniel-thom something happened that the run with main gets cancelled and I don't know why. Haven't had a chance to figure it out.

jd-lara (approved):

PR #41, opened 2023-04-27
New methods for LODF: improvements and virtual LODF

alefcastelli:
Faster LODF computation based on sparse matrix, new Virtual LODF methods similar to Virtual PTDF


src/virtual_lodf_calculations.jl
```diff
@@ -0,0 +1,282 @@
...
+    axes::Ax
+    lookup::L
+    valid_ix::Vector{Int}
+    temp_data::Vector{Float64}
+    cache::RowCache
+    subnetworks::Dict{Int, Set{Int}}
+    tol::Base.RefValue{Float64}
+end
+
+function _get_PTDF_A_diag(
```
jd-lara, line 50: This code is duplicated. We probably can get away with making a single function for this

jd-lara (approved):

jd-lara:
The results of loading the LODF in the MMWG dataset are 227.657088 seconds (1.71 M allocations: 193.263 GiB, 2.61% gc time, 0.01% compilation time) which is reasonable enough for now. More improvements later.

PR #50, opened 2023-08-24
simple updated to make isempty workable for each data structures

alefcastelli:
Without theses changes, PowerFlows.jl was throwing an error related to `Base.show`. The following changes fix the issue.


jd-lara:
@alefcastelli can you post the error. There is no reason for PF's to be throwing an error because of show methods.

jd-lara (approved):

PR #54, opened 2023-09-06
Jd/psy3

jd-lara:
This PR requires the corresponding branches in PSY 3


src/virtual_lodf_calculations.jl
```diff
@@ -101,7 +101,8 @@ end
 
 function VirtualLODF(
     branches,
-    buses::Vector{PSY.Bus};
+    buses::Vector{PSY.ACBus};
+    dist_slack::Vector{Float64} = Float64[],
```
alefcastelli, line 105: no need of dist_slack

alefcastelli (changes requested):
VirtualLODF does not need dist_slack argument

alefcastelli:
Ready to merge. Tests still to pass and must be done as soon as PowerSystems 3.0 is released.

PR #66, opened 2023-12-20
fix ptdf constructor

### LINKED ISSUE #61: PTDF and LODF elements are NaNs
In case the loaded system contains branches with null impedance, its inverse is Inf.
Therefore, the BA and ABA matrices (and its LU factorization ones) contain Inf values. When trying to solve the linear system the resulting matrix (PTDF or LODF) will have all elements equal to NaNs.

Actions must be taken to avoid this situation since it can be very hard for a user to understand where the issue could be.

jd-lara:
closes #61 also


jd-lara:
closed, these changes are included in #67

PR #67, opened 2023-12-22
afc/radial_branches_matrices

alefcastelli:
The PR adds new features regarding the possibility to evaluate all the network matrices considering a simplified system without leaf buses and radial branches.

The option `reduce_radial_branches` is available for all matrix types.


jd-lara:
@alefcastelli I made some improvements to the RadialBranches code for the integration with PowerSimulations.jl. One test for the PTDF seems to be failing but it isn't clear why. I think that the test is written in a way that the indexing doesn't work as expected.

src/lodf_calculations.jl
```diff
@@ -413,13 +442,25 @@ function LODF(
     validate_linear_solver(linear_solver)
     ax_ref = make_ax_ref(A.axes[1])
     if reduce_radial_branches
-        radial_branches = RadialBranches(A)
+        if !isempty(BA.radial_branches) && !isempty(ABA.radial_branches)
```
jd-lara, line 445: we need to check that the radial branches object is in fact the same

src/ptdf_calculations.jl
```diff
@@ -439,15 +429,33 @@ Builds the PTDF matrix from a system. The return is a PTDF array indexed with th
...
 # Arguments
 - `sys::PSY.System`:
-        Power Systems system
+        PSY system for which the matrix is constructed
+
+# Keyword Arguments
+- `dist_slack::Vector{Float64}=Float64[]`:
+        vector of weights to be used as distributed slack bus.
+        The distributed slack vector has to be the same length as the number of buse
+- `reduce_radial_branches::Bool=false`:
```
jd-lara, line 438: kwarg inconsistency again

src/ptdf_calculations.jl
```diff
@@ -439,15 +429,33 @@ Builds the PTDF matrix from a system. The return is a PTDF array indexed with th
...
+        if True the matrix will be evaluated discarding
+        all the radial branches and leaf buses (optional, default value is false)
+- `kwargs...`:
```
jd-lara, line 441: This is not needed

src/virtual_ptdf_calculations.jl
```diff
@@ -157,13 +156,42 @@ end
...
+        PSY system for which the matrix is constructed
+
+# Keyword Arguments
+- `dist_slack::Vector{Float64}=Float64[]`:
+        vector of weights to be used as distributed slack bus.
+        The distributed slack vector has to be the same length as the number of buse
+- `reduce_radial_branches::Bool=false`:
+        if True the matrix will be evaluated discarding
+        all the radial branches and leaf buses (optional, default value is false)
+- `kwargs...`:
```
jd-lara, line 171: not needed

src/virtual_ptdf_calculations.jl
```diff
@@ -157,13 +156,42 @@ end
...
+
+# Arguments
+- `sys::PSY.System`:
+        PSY system for which the matrix is constructed
+
+# Keyword Arguments
+- `dist_slack::Vector{Float64}=Float64[]`:
+        vector of weights to be used as distributed slack bus.
+        The distributed slack vector has to be the same length as the number of buse
+- `reduce_radial_branches::Bool=false`:
```
jd-lara, line 168: Inconsistency in keyword argument

src/BA_ABA_matrices.jl
```diff
@@ -1,9 +1,10 @@
+# TODO: BA_Matrix(sys::PSY.Sytems, radial_branches::Vector{String}), for custom radial branches
```
jd-lara, line 1: Is this still a TODO?

src/ptdf_calculations.jl
alefcastelli: insonsistency due to the need to keep dist_slack and reduce_radial_branches explicit

src/lodf_calculations.jl
alefcastelli: done

PR #69, opened 2024-01-15
PSI integration part 2

jd-lara:
Changes to comply with PSI PR review.

### DIFF
--- a/src/BA_ABA_matrices.jl
+++ b/src/BA_ABA_matrices.jl
@@ -15,8 +15,8 @@ Structure containing the BA matrix and other relevant data.
         with the names of buses and branches
 - `ref_bus_positions::Set{Int}`:
         Set containing the indexes of the columns of the BA matrix corresponding
-        to the refence buses
-- `radial_branches::RadialNetworkReduction`:
+        to the reference buses
+- `radial_network_reduction::RadialNetworkReduction`:
         Structure containing the radial branches and leaf buses that were removed
         while evaluating the matrix
 """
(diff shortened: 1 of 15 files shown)


src/BA_ABA_matrices.jl
```diff
@@ -16,7 +16,7 @@ Structure containing the BA matrix and other relevant data.
 - `ref_bus_positions::Set{Int}`:
         Set containing the indexes of the columns of the BA matrix corresponding
         to the refence buses
```
GabrielKS, line 18:
```suggestion
        to the reference buses
```

src/BA_ABA_matrices.jl
```diff
@@ -16,7 +16,7 @@ Structure containing the BA matrix and other relevant data.
...
-- `radial_branches::RadialBranches`:
+- `radial_branches::RadialNetworkReduction`:
```
GabrielKS, line 19: To complete the renaming we should probably change all the field names too, right? Unless domain knowledge makes this non-confusing.
```suggestion
- `radial_network_reduction::RadialNetworkReduction`:
```

src/BA_ABA_matrices.jl
```diff
@@ -25,7 +25,7 @@ struct BA_Matrix{Ax, L <: NTuple{2, Dict}} <: PowerNetworkMatrix{Float64}
     axes::Ax
     lookup::L
     ref_bus_positions::Set{Int}
-    radial_branches::RadialBranches
+    radial_branches::RadialNetworkReduction
```
GabrielKS, line 28: (see above)

GabrielKS (changes requested):
Looks good aside from the typo and the suggestion to rename `radial_branches` and `rb` fields wherever they appear unless domain knowledge makes this non-confusing.

GabrielKS (approved):
Looks good!

PR #75, opened 2024-02-25
Update Project.toml

jd-lara:
Testing bump in Pardiso to avoid issues with versions


jd-lara:
I tested locally on my M1 Mac that this solution works

PR #89, opened 2024-09-21
Jd/index fixes

### LINKED ISSUE #86: Adjacency function UndefVarError
When running "PNM.Adjacency(system)", I get the error "UndefVarError: `_make_ax_ref` not defined"

jd-lara:
closes #86


rodrigomha (approved):
Looks good, but not sure if the change to make_ax_ref is working as it is not being tested

PR #95, opened 2025-01-14
add Yft, Ytf to power_network_matrix Ybus

rbolgaryn:


rbolgaryn:
Add branch admittance matrices (Yft, Ytf) to the Ybus structure so that we can efficiently calculate branch flows in PowerFlows in AC power flow calculation.

src/ybus_calculations.jl
```diff
@@ -7,16 +7,27 @@ struct Ybus{Ax, L <: NTuple{2, Dict}} <: PowerNetworkMatrix{ComplexF64}
     data::SparseArrays.SparseMatrixCSC{ComplexF64, Int}
     axes::Ax
     lookup::L
+    data_ft::SparseArrays.SparseMatrixCSC{ComplexF64, Int}
```
GabrielKS, line 10: Edit docstring to describe how this also represents line data

src/ybus_calculations.jl
```diff
@@ -7,16 +7,27 @@ struct Ybus{Ax, L <: NTuple{2, Dict}} <: PowerNetworkMatrix{ComplexF64}
...
+    data_tf::SparseArrays.SparseMatrixCSC{ComplexF64, Int}
+    fb::Vector{Int64}
```
GabrielKS, line 12: To be as flexible as the existing struct these would type parameters:
```suggestion
    fb::Ax_fb
    tb::Ax_tb
```

though I think the existing struct is probably under-constrained, maybe it should be something like
```julia
struct Ybus{T} <: PowerNetworkMatrix{ComplexF64}
    data::SparseArrays.SparseMatrixCSC{ComplexF64, Int}
    axes::Tuple{Vector{T}, Vector{T}}
    lookup::Tuple{Dict{T, Int}, Dict{T, Int}}
    data_ft::SparseArrays.SparseMatrixCSC{ComplexF64, Int}
    data_tf::SparseArrays.SparseMatrixCSC{ComplexF64, Int}
    fb::Vector{T}
    tb::Vector{T}
end
```
but that would break backwards compatibility.

src/ybus_calculations.jl
```diff
@@ -7,16 +7,27 @@ struct Ybus{Ax, L <: NTuple{2, Dict}} <: PowerNetworkMatrix{ComplexF64}
     data::SparseArrays.SparseMatrixCSC{ComplexF64, Int}
     axes::Ax
     lookup::L
+    data_ft::SparseArrays.SparseMatrixCSC{ComplexF64, Int}
+    data_tf::SparseArrays.SparseMatrixCSC{ComplexF64, Int}
```
GabrielKS, line 11: Do these `data_ft` and `data_tf` matrices always have just one nonzero entry per row?

GabrielKS (changes requested):
A few minor things. I might not yet fully understand the use case. Also, this is missing tests.

src/ybus_calculations.jl
GabrielKS: Resolved.

src/ybus_calculations.jl
GabrielKS: Resolved.

jd-lara (changes requested):
These additional entries are in fact different Matrices. They should be separated into their own object and potentially store just one matrix but implement methods that allow the calculations to be done efficiently.

rbolgaryn:
This PR should be good now

src/PowerNetworkMatrices.jl
```diff
@@ -20,6 +20,7 @@ export validate_connectivity
 export VirtualLODF
 export VirtualPTDF
 export Ybus
+export _buildybus
```
jd-lara, line 23: we don't export internal methods.

rbolgaryn:
todo: 1) make also the ft, tb, sb, nullable, 2) make the Ybus structure to be "kwdef"

src/ybus_calculations.jl
```diff
@@ -2,55 +2,85 @@
 Nodal admittance matrix (Ybus) is an N x N matrix describing a power system with N buses. It represents the nodal admittance of the buses in a power system.
 
 The Ybus Struct is indexed using the Bus Numbers, no need for them to be sequential
+
+The fields yft and ytf are the branch admittance matrices for the from-to and to-from branch admittances respectively. They are indexed with the branch numbers and the bus numbers.
+The branch numbers are sequential. The bus numbers are represented by fb, tb arraus of sequential bus indices (internal bus indices).
```
GabrielKS, line 7:
```suggestion
The branch numbers are sequential. The bus numbers are represented by fb, tb arrays of sequential bus indices (internal bus indices). 
```

GabrielKS (commented):
Deferring to @jd-lara

src/ybus_calculations.jl
```diff
@@ -2,55 +2,85 @@
 Nodal admittance matrix (Ybus) is an N x N matrix describing a power system with N buses. It represents the nodal admittance of the buses in a power system.
 
 The Ybus Struct is indexed using the Bus Numbers, no need for them to be sequential
+
+The fields yft and ytf are the branch admittance matrices for the from-to and to-from branch admittances respectively. They are indexed with the branch numbers and the bus numbers.
+The branch numbers are sequential. The bus numbers are represented by fb, tb arrays of sequential bus indices (internal bus indices).
```
jd-lara, line 7: there are no branch numbers in PSY. We index with the branch names like in the Ybus

jd-lara (approved):
There is one small change to consider about the branch indexing

jd-lara (approved):

PR #98, opened 2025-02-04
Add ward reduction 

m-bossart:
Outstanding issues:

- [x] Accounting for virtual admittances in various `PowerNetworkMatrix` types
- [x] Reassigning reference bus
- [x] Improving performance of Z matrix computation
- [x] Add tests for PTDF with ward reduced system


m-bossart:
Items to discuss:

- [ ] This PR also adds a check during radial reduction to not radially reduce small islanded areas (<10 buses). To accommodate this change, the API for doing the radial reduction is limited to the method that takes the full system.
- [ ] The current implementation of ward inverts the Ybus using KLU. I'd like to discuss the need/details for an alternative implementation for larger systems. 
- [ ] The tests largely focus on the proper axes for various matrices when using the ward reduction. Ideally we can include more tests for the correctness of the ward reduction (i.e the parameters associated with the additional branches and admittances). 
- [ ] Discuss the desired behavior or error for corner cases.

src/definitions.jl
```diff
@@ -7,3 +7,5 @@ const MAX_CACHE_SIZE_MiB = 100
 const ROW_PERSISTENT_CACHE_WARN = 1 * GiB
 
 DEFAULT_LODF_CHUNK_SIZE = 18_000
+
+IS.@scoped_enum(NetworkReductionTypes, RADIAL = 1, WARD = 2, RADIAL_AND_WARD = 3)
```
jd-lara, line 11:
```suggestion
IS.@scoped_enum(NetworkReductionTypes, RADIAL = 1, WARD = 2, RADIAL_THEN_WARD = 3) # Performing WARD and then RADIAL isn't supported
```

m-bossart:
Changed so that the user specifies buses that should be exempt from reduction (the 10 bus limit seems very arbitrary)

Project.toml
```diff
@@ -4,22 +4,24 @@ authors = ["Jose Daniel Lara", "Alessandro Francesco Castelli", "Sourabh Dalvi"]
 version = "0.12.1"
 
 [deps]
+AppleAccelerate = "13e28ba4-7ad8-5781-acae-3021b1ed3924"
```
jd-lara, line 7: There is no need to add this as a dependency

jd-lara: It might be an oversight from previous code

jd-lara (commented):
There are changes to the branch hierarchy for PowerSystems 5 that might be relevant here.

Project.toml
m-bossart: So should I remove this then? https://github.com/NREL-Sienna/PowerNetworkMatrices.jl/blob/8ce328926dc1e862a39a114a29fcea61a2c5f0b3/src/PowerNetworkMatrices.jl#L41

m-bossart:
> @m-bossart is the code here PSY5 ready? If it is, we need to change the target from main to PSY5.

done

m-bossart:
@jd-lara Should we rename `get_ac_branches` to `get_ac_transmission`? To be consistent with the new type hierarchy? 
https://github.com/NREL-Sienna/PowerNetworkMatrices.jl/blob/8ce328926dc1e862a39a114a29fcea61a2c5f0b3/src/common.jl#L16

jd-lara (approved):
I am going to merge this to PSY5 because it is getting very unwieldy

Issue #102, opened 2025-03-19
`getindex` on `ABA_Matrix`


luke-kiernan:
I think there's a few bugs lurking in `getindex` on `ABA_Matrix`.
```
# NOTE: bus_1, bus_2 are bus numbers not row and column indices!
function Base.getindex(A::ABA_Matrix, bus_1, bus_2)
    if bus_1 in A.ref_bus_positions || bus_2 in A.ref_bus_positions
          ...error message about "ABA matrix doesn't have slack bus"...
```
If bus_1 and bus_2 are bus numbers, not indices, then looking them up in `A.ref_bus_positions` (which stores indices) doesn't make sense. Further, this doesn't seem to be the only issue. The following throws a `BoundsError` when looking up the pair of buses `(325, 101)`, yet neither of those are slack buses:
```
using PowerNetworkMatrices
using PowerSystemCaseBuilder
const PSB = PowerSystemCaseBuilder
const PNM = PowerNetworkMatrices
sys = build_system(PSISystems, "RTS_GMLC_DA_sys")
M = ABA_Matrix(sys)
for I in keys(M); M[I]; end;
```
And the following should error, but doesn't (as mentioned above):
```
slackBus = [bus for bus in PSY.get_components(PSY.ACBus, sys) if get_bustype(bus) == PSY.ACBusTypes.REF][1]
M[slackBus.number, slackBus.number]
```

jd-lara:
If bus_1 and bus_2 are buses not numbers this implementation of `getindex`

PR #105, opened 2025-03-26
Fix issue #102

luke-kiernan:
I added a `ref_bus_numbers::Set{Int}` field to `ABA_Matrix`, and replaced `ref_bus_positions` with `ref_bus_numbers` in the spots mentioned in issue #102. I also changed the system in the test case to`RTS_GMLC_DA_sys`, one where the bus numbers aren't consecutive integers. (Lines 3-37 in `test_BA_ABA_matrix.jl` could be reverted.)

### DIFF
--- a/test/test_BA_ABA_matrix.jl
+++ b/test/test_BA_ABA_matrix.jl
@@ -1,44 +1,44 @@
 @testset "Test A, BA, and ABA matrix creation" begin
     # test on 5 and 14 bus system
-    for name in ["c_sys5", "c_sys14"]
-        sys = PSB.build_system(PSB.PSITestSystems, name)
-        # at first let's see if factorization flag works
-        ABA_no_lu = ABA_Matrix(sys)
-        @test isnothing(ABA_no_lu.K)
-        # check the is_factorized function
-        @test is_factorized(ABA_no_lu) == false
-        # factorize the ABA matrix
-        ABA_no_lu = factorize(ABA_no_lu)
-        @test is_factorized(ABA_no_lu) == true
-        # get the ABA matrix with the current method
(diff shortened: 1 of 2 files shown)


luke-kiernan:
I don't have write access, so I'm issuing the PR from a fork. Also, the failed performance check seems to be random variation: when I run `@benchmark` on the two, the output is comparable.
Screenshot 2025-03-27 at 4 03 49 PM
Screenshot 2025-03-27 at 4 23 10 PM

test/test_BA_ABA_matrix.jl
```diff
@@ -1,44 +1,44 @@
...
-        A = IncidenceMatrix(sys)
-        BA = BA_Matrix(sys)
-        ABA_2 = transpose(A.data) * BA.data'
-        @test isapprox(
-            ABA_lu.data,
-            ABA_2[setdiff(1:end, A.ref_bus_positions), setdiff(1:end, A.ref_bus_positions)],
-            atol = 1e-8,
-        )
+    #for name in ["c_sys5", "c_sys14"]
+    sys = PSB.build_system(PSB.PSISystems, "RTS_GMLC_DA_sys")
```
jd-lara, line 4: Why remove the other test systems here?

jd-lara (commented):
overall looks good to me, needs to address the comment on the testing

test/test_BA_ABA_matrix.jl
luke-kiernan: I made that change while tracking down the bug, then forgot to undo it before pushing to the remote. Fixed now.

jd-lara (approved):

PR #112, opened 2025-04-22
perf: sizehint and push in get_buses

luke-kiernan:
Buses are mutable: making a vector of undefined references doesn't really gain us much here. Instead, use `sizehint!` and push to the end. (Discovered while profiling `PowerFlows.jl`.)

### DIFF
--- a/src/common.jl
+++ b/src/common.jl
@@ -57,9 +57,9 @@ function get_buses(
         end
     end
 
-    count_i = 1
     all_buses = PSY.get_components(PSY.ACBus, sys)
-    buses = Vector{PSY.ACBus}(undef, length(all_buses))
+    buses = Vector{PSY.ACBus}()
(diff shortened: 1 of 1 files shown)


luke-kiernan:
Huh. Benchmarking it as follows, the sizehint-and-push strategy seems comparable to the current code.
```
sys = PSB.build_system(MatpowerTestSystems, "matpower_ACTIVSg10k_sys")
empty_reduction = Dict{Int64, Set{Int64}}()
PNM.get_buses(sys, empty_reduction)
@benchmark PNM.get_buses($sys, $empty_reduction) seconds = 10
```
(I expected it to be better.) However, it's awkward to benchmark, due to the outsized impact of GC. My other reason for prefering this version: it avoids undefined references--more type stable.

luke-kiernan:
As mentioned above, this doesn't really help performance, so I'll just close it.

PR #115, opened 2025-04-28
patch reversemap bus_indices

rodrigomha:


jd-lara (commented):
lets merge this to avoid broken capabilities but we should make some of this more systematic on the mapping

PR #116, opened 2025-04-28
Mb/switch reduction

m-bossart:
Fixes bad merge and makes tests pass (makes 3WT transformer and network reduction changes compatible) 

 disables the new network reduction tests


jd-lara (approved):
lets merge this to avoid broken capabilities but we should make some of this more systematic on the mapping

PR #119, opened 2025-05-12
Mb/reductions refactor

m-bossart:
So far computes the Ybus with "default" reductions. 

Passes `test_ybus` and `test_ybus_psse`. Other tests for other matrices are not passing.


m-bossart:
@jd-lara  Here is the starting point to implement the other matrices and check that the `NetworkReduction` is sufficient

Issue #122, opened 2025-05-19
Testing for unavailable components


m-bossart:
Should add test coverage for the impact of unavailable components (throughout the package)

PR #123, opened 2025-05-19
fix constructor and handling unavailable branches

### LINKED ISSUE #121: Ybus constructor errors
The `Ybus` constructor [here](https://github.com/NREL-Sienna/PowerNetworkMatrices.jl/blob/mb/reductions-refactor/src/ybus_calculations.jl#L567) is missing the `make_branch_admittances` argument. Similarly, `return Ybus(ybus, adj, axes, lookup, nr)` at the end of that function errors: the 4 fields whose values depend on `make_branch_admittances` (`yft`, `ytf`, `fb`, and `tb`) are missing.

m-bossart:
closes #121 
adding a separate issue to improve the testing for unavailable components #122


luke-kiernan (approved):
Looks good--I ran the tests that in PowerFlows.jl that brought my attention to the issue and they pass now.

PR #129, opened 2025-06-30
Find subnetworks via iterative union-find

luke-kiernan:
Use an iterative union-find algorithm in `find_subnetworks`, instead of a depth-first search. (Memory usage should be similar: before we had a set of size n, whereas now we have a dictionary of size n.)


src/union_find_struct.jl
```diff
@@ -0,0 +1,22 @@
+struct UnionFind
+    rep::Dict{Int, Int} # maps each element to its representative
+end
+
+UnionFind(elems::Vector{Int}) = UnionFind(Dict{Int, Int}(x => x for x in elems))
+
+function get_rep(uf::UnionFind, x::Int)
```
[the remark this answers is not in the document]
rbolgaryn: do you think it should have ! like the get! for Dict?

src/union_find_struct.jl
```diff
@@ -0,0 +1,22 @@
...
+    while uf.rep[x] != x
```
rbolgaryn, line 8: I wonder if this reduction of uf.rep[x] by 1 makes any difference at all:

```Julia
while (r = uf.ref[x]) != x
    x = uf.ref[x] = uf.ref[r]
end
```

Or allocating r is worse than querying the Dict one more time.

A shame this wouldn't work: x = r = uf.ref[r]

src/union_find_struct.jl
```diff
@@ -0,0 +1,22 @@
...
+        uf.rep[x] = uf.rep[uf.rep[x]] # path compression
+        x = uf.rep[x]
+    end
+    return x
+end
+
+function union!(uf::UnionFind, x::Int, y::Int)
+    rootX = get_rep(uf, x)
+    rootY = get_rep(uf, y)
+    if rootX != rootY
```
rbolgaryn, line 18: so here you are re-using the bus number to also represent a rank? what do you think about assigning a new_root first as min(...) and then using it to avoid calling min twice?

test/test_adjacency.jl
```diff
@@ -13,3 +13,27 @@
...
+    A = SparseArrays.sparse(I(n))
+    for (i, j) in edge_inds
+        A[i, j] = 1
+        A[j, i] = 1
+    end
+    test_subnetworks = PNM.find_subnetworks(A, buses)
+    expected =
+        [Set(100 .+ (1:3)), Set(100 .+ (4:8)), Set(100 .+ (9:9)), Set(100 .+ (10:11))]
+    @test length(values(test_subnetworks)) == length(expected)
+    for (k, v) in test_subnetworks
```
rbolgaryn, line 33: because you have @test in a loop here I suggest adding failfast=true to the @testset above to reduce spamming of the console

luke-kiernan: Sure. It's a loop over just 4 things, but I suppose there's no harm in adding it.

jd-lara (commented):
I have never seen this method so I need to understand it better before merge.

PR #130, opened 2025-07-01
fix get bus index/indices calls; variable for path to test_data

luke-kiernan:


luke-kiernan:
I changed the destination branch and removed the indexing changes (since the new destination branch fixes that already).

PR #136, opened 2025-07-08
Check arc validity

luke-kiernan:


src/PowerNetworkMatrix.jl
```diff
@@ -321,7 +321,7 @@ function get_branch_multiplier(A::T, branch_name::String) where {T <: PowerNetwo
         if branch_name == PSY.get_name(k)
             parallel_branch_set = nr.parallel_branch_map[v]
             multiplier = _compute_parallel_multiplier(parallel_branch_set, branch_name)
-            @error "multiplier for entry corresponding to parallel branch: $multiplier"
+            # @error "multiplier for entry corresponding to parallel branch: $multiplier"
```
m-bossart, line 324: Just remove this please

src/common.jl
```diff
@@ -69,20 +88,8 @@ function get_ac_branches(
...
-        end
-        if PSY.get_bustype(arc.to) == ACBusTypes.ISOLATED
-            throw(
-                IS.ConflictingInputsError(
-                    "Branch $(PSY.get_name(br)) is set available and connected to isolated bus $(PSY.get_name(arc.to))",
-                ),
-            )
-        end
+        name = PSY.get_name(br)
+        check_arc_validity(arc, name)
```
m-bossart, line 92: `get_ac_branches` will not be used once the refactoring is complete

m-bossart (commented):
What system do you have that is failing this check? This could indicate an issue with the parsing because we shouldn't get the case where we have available components connected to isolated buses.

luke-kiernan:
> What system do you have that is failing this check? This could indicate an issue with the parsing because we shouldn't get the case where we have available components connected to isolated buses.

It was a test that was marked as a TODO: it's specifically adding an isolated bus and a branch, then testing if an error is thrown. Before, none was thrown, so I added one.

src/common.jl
```diff
@@ -55,6 +56,24 @@ function get_bus_indices(
     return fr_bus_ix, to_bus_ix
 end
 
+function check_arc_validity(arc::PSY.Arc, name::String)
+    if PSY.get_bustype(arc.from) == ACBusTypes.ISOLATED
```
rbolgaryn, line 60: I think we should be using `get_from(arc)`

PR #142, opened 2025-07-10
Jd/2d reduction

jd-lara:
Openning this PR for discussion on the changes needed for the 2d reduction


src/AdjacencyMatrix.jl
jd-lara, line 1: @m-bossart I don't get why we removed this object...

src/degree_two_reduction.jl
```diff
@@ -1,13 +1,22 @@
...
     sys::PSY.System,
     ::Val{NetworkReductionTypes.DEGREE_TWO},
 )
+    irreducible_buses = Set{Int}()
+    for c in get_components(StaticInjection, sys_DA)
+        push!(irreducible_buses, PSY.get_number(PSY.get_bus(c)))
+    end
+    irreducible_indices = Set([yb.lookup[2][i] for i in irreducible_buses])
+
+    arc_maps = find_degree2_chains(A.data, irreducible_indices)
```
jd-lara, line 13: The complete logic isn't implemented yet to connect back the reduction data to the branches, but this code will find all the degree-2 reductions needed.

src/degree_two_reduction.jl
jd-lara, line 1: @luke-kiernan you might find this code interesting. It is my recursive implementation to find chains of degree two nodes and find the corresponding new arcs that "absorb" the unnecessary arcs.

src/connectivity_checks.jl
jd-lara, line 1: This file needs to be cleaned up.

src/degree_two_reduction.jl
```diff
@@ -1,13 +1,22 @@
-#NOTE: hardcoded for testing system 
+#NOTE: hardcoded for testing system
 function get_reduction(
-    A::IncidenceMatrix,
+    A::AdjacencyMatrix,
     sys::PSY.System,
     ::Val{NetworkReductionTypes.DEGREE_TWO},
 )
+    irreducible_buses = Set{Int}()
```
jd-lara, line 7: Note to myself, this selection can consider active or reactive power. so that we can eliminate buses with reactive power support devices.

src/degree_two_reduction.jl
jd-lara: In this format prev_node can take nothing and Int.

luke-kiernan: That's fixable: replace with `-1` (if node numbers are always positive) or `current_node` (if nodes aren't neighbors of themselves). If you want me to write up a fully working, type stable, iterative version, I can do that.

m-bossart (approved):

PR #156, opened 2025-07-21
Add kwarg to Ybus whether to include the constant impedance loads as shunt impedances in Ybus

rbolgaryn:


m-bossart (commented):
Even though many of the psse tests don't have constant impedance loads so will pass with or without the keyword, my preference is to include it as true for all those tests so it is clear it is needed if the goal is to match PSSe and there are constant impedance loads.

jd-lara (approved):

PR #159, opened 2025-07-22
Finish degree two and test all reductions

### LINKED ISSUE #113: Add Degree-2 No injection branch/bus reduction
As we continue the path to reduce system size without sacrificing power flow results. We should incorporate degree-2 no injection reduction. 

This will help up remove additional injection buses and branch expressions from the PTDF.

### LINKED ISSUE #146: Additional testing for new reductions
- Add tests for correctness of reductions (DegreeTwo, Radial, Ward).
- Add tests for multiple reductions (i.e. making sure the reductions compose properly). 
- These tests should test all fields of `NetworkReductionData`
- Will need to develop a small test system that has all reduction types and can replicate the patterns seen in larger datasets.

m-bossart:
Adds testing for correctness of each reduction (default ones, `RadialReduction`, `DegreeTwoReduction`, `WardReduction`)
Finishes the connection from the core algorithm to the full implementation for the `DegreeTwoReduction`
All the tests are passing.
With this PR, can run the `RadialReduction` and `DegreeTwoReduction` on both EI and WECC
closes #113 
closes #146 

@jd-lara I made a few changes/fixes to the d2 algorithm when developing the tests:

1. Fix to consider the irreducible buses when deciding to terminate a chain.
2. Modification to consider subchains if the arc for the full chain is already included.

### DIFF
--- a/src/Ybus.jl
+++ b/src/Ybus.jl
@@ -1077,6 +1077,7 @@ function _remake_reverse_direct_branch_map!(nr::NetworkReductionData)
         reverse_direct_branch_map[v] = k
     end
     nr.reverse_direct_branch_map = reverse_direct_branch_map
+    return
 end
 function _remake_reverse_parallel_branch_map!(nr::NetworkReductionData)
     reverse_parallel_branch_map = Dict{PSY.Branch, Tuple{Int, Int}}()
@@ -1086,6 +1087,7 @@ function _remake_reverse_parallel_branch_map!(nr::NetworkReductionData)
         end
     end
     nr.reverse_parallel_branch_map = reverse_parallel_branch_map
+    return
 end
 function _remake_reverse_series_branch_map!(nr::NetworkReductionData)
     reverse_series_branch_map = Dict{PSY.Branch, Tuple{Int, Int}}()
@@ -1095,7 +1097,9 @@ function _remake_reverse_series_branch_map!(nr::NetworkReductionData)
         end
     end
     nr.reverse_series_branch_map = reverse_series_branch_map
+    return
 end
+
 function _remake_reverse_transformer3W_map!(nr::NetworkReductionData)
     reverse_transformer3W_map =
         Dict{Tuple{PSY.ThreeWindingTransformer, Int}, Tuple{Int, Int}}()
@@ -1103,6 +1107,7 @@ function _remake_reverse_transformer3W_map!(nr::NetworkReductionData)
         reverse_transformer3W_map[v] = k
     end
     nr.reverse_transformer3W_map = reverse_transformer3W_map
+    return
 end
 
 """
@@ -1174,6 +1179,7 @@ function _validate_study_buses(
             end
         end
     end
+    return
 end
 
 function get_reduction(


src/common.jl
```diff
@@ -230,169 +271,40 @@ function validate_linear_solver(linear_solver::String)
...
-        push!(BA_V, b_val)
-
-        push!(BA_I, to_b)
-        push!(BA_J, ix + iix - 1)
-        push!(BA_V, -b_val)
+function validate_buses(ybus::Ybus, buses::Vector{Int})
+    reverse_bus_search_map = ybus.network_reduction_data.reverse_bus_search_map
+    for bus_no in buses
+        reduced_bus_no = get(reverse_bus_search_map, bus_no, bus_no)
+        if reduced_bus_no ∉ ybus.axes[1]
```
jd-lara, line 282: For clarity can you make a getter in the Ybus to grab this axes. We have this pattern all over but it isn't immediately obvious what we are doing.

src/degree_two_reduction.jl
```diff
@@ -300,11 +325,38 @@ function find_degree2_chains(
...
+        @warn "Nodes $(chain_path[1]) and $(chain_path[end]) already have a parallel path, searching for valid subchains."
+        subchains = Vector{Int}[]
+        n = length(chain_path)
+        for i in 1:n
+            for j in i:n
+                push!(subchains, chain_path[i:j])
+            end
+        end
+        sort!(subchains; by = length, rev = true)
+        filter!(x -> length(x) > 2, subchains)
```
jd-lara, line 352: why this changes? doing a sort! and a filter with an anonymous function in a hot loop isn't a good idea. We need to discuss this one

src/ward_reduction.jl
```diff
@@ -167,7 +137,7 @@ function _validate_study_buses(
         for (_, v) in sub_networks
```
jd-lara, line 137:
```suggestion
        for v in values(sub_networks)
```

src/ybus_calculations.jl
```diff
@@ -943,9 +921,24 @@ function _apply_reduction(ybus::Ybus, nr_new::NetworkReductionData)
     #Assumes only the last reduction (Ward) can add branches and admittances
     nr.added_branch_map = nr_new.added_branch_map
     nr.added_admittance_map = nr_new.added_admittance_map
-
-    #TODO - loop through added branches and admittances and modify the Ybus for Ward reduction
+    if isempty(nr.series_branch_map)
+        nr.series_branch_map = nr_new.series_branch_map
+        nr.reverse_series_branch_map = nr_new.reverse_series_branch_map
+    elseif !isempty(nr_new.series_branch_map) && !!isempty(nr.series_branch_map)
```
jd-lara, line 927: is there a double negation here?

src/ybus_calculations.jl
```diff
@@ -980,6 +973,83 @@ function _apply_reduction(ybus::Ybus, nr_new::NetworkReductionData)
...
+        end
+        data[bus_lookup[equivalent_arc[1]], bus_lookup[equivalent_arc[2]]] +=
+            1 / series_impedance_from_to
+        data[bus_lookup[equivalent_arc[2]], bus_lookup[equivalent_arc[1]]] +=
+            1 / series_impedance_to_from
+        data[bus_lookup[equivalent_arc[1]], bus_lookup[equivalent_arc[1]]] +=
+            shunt_admittance / 2
+        data[bus_lookup[equivalent_arc[2]], bus_lookup[equivalent_arc[2]]] +=
+            shunt_admittance / 2
+    end
```
jd-lara, line 1004:
```suggestion
    return
    end
```

test/test_ybus_reductions.jl
```diff
@@ -18,46 +18,196 @@
     )
 end
 
-@testset "Basic degree two NetworkReductionData" begin
+function check_bus_arc_axis_consistency(A::IncidenceMatrix)
+    arc_axis_numbers = Set()
+    for arc in A.axes[1]
```
jd-lara, line 23: another place where a getter is useful

jd-lara (changes requested):
We need to do a couple of changes here.

src/degree_two_reduction.jl
jd-lara: something like this would be internally optimized by Julia for not huge chains 
```
sort([x for x in data if length(x) > 2], by = length, rev = true)
```

src/network_reduction.jl
```diff
@@ -14,8 +14,8 @@ abstract type NetworkReduction end
         Dict{PSY.Branch, Tuple{Int, Int}}()
     series_branch_map::Dict{Tuple{Int, Int}, Vector{Any}} =
         Dict{Tuple{Int, Int}, Vector{Any}}()
-    reverse_series_branch_map::Dict{PSY.Branch, Tuple{Int, Int}} =
-        Dict{PSY.Branch, Tuple{Int, Int}}()
+    reverse_series_branch_map::Dict{Any, Tuple{Int, Int}} =
+        Dict{Any, Tuple{Int, Int}}()
```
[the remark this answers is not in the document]
m-bossart: Will track this here: https://github.com/NREL-Sienna/PowerNetworkMatrices.jl/issues/162

src/ybus_calculations.jl
m-bossart: there shouldn't be, will fix this

m-bossart:
Just two tests failing due to : https://github.com/NREL-Sienna/PowerSystemCaseBuilder.jl/issues/149

jd-lara (approved):

jd-lara (approved):

PR #160, opened 2025-07-23
Add get _retained_branches_names

m-bossart:
This is a helper function to help restore functionality to PSI. The function returns the branch names that are not in parallel with other branches. Does not include three winding transformer branches.


jd-lara (approved):
Let's merge this into the refactor of the mappings.

PR #161, opened 2025-07-24
add transformer phase shift

rbolgaryn:
Adding transformer phase shift to Transformer2W and TapTransformer as `phase_shift`


jd-lara (commented):
Can we test of this makes any changes in EI and WECC before we merge?

PR #173, opened 2025-08-07
include line conductance in ybus calculation

luke-kiernan:


m-bossart (commented):
The last change for subnetwork line was just fixed in psy5 branch. I use the bus_ax instead of the keys(bus_lookup). Please remove that change to fix the conflict

luke-kiernan:
> Please remove that change to fix the conflict

Should be good now.

PR #177, opened 2025-08-11
refactor ybus code

luke-kiernan:
Replace some of the copy-pastes in computing the ybus entries with multiple dispatch. Some of this infrastructure (`_get_angle`, `_get_tap`, `TransformerWindingDispatch`) could be moved into PSY if desired.

(In order to recover the flows after a degree 2 reduction, I need to go from the component to the the 4 `y_{ff/tf/ft/tt}` numbers. Since I was refactoring things anyway, I decided to take the time to figure out how to de-duplicate things.)


src/Ybus.jl
```diff
@@ -202,349 +229,121 @@ function _ybus!(
...
-    tb::Vector{Int},
-    ix::Int,
-    nr::NetworkReductionData,
-    adj::SparseArrays.SparseMatrixCSC{Int8, Int},
+```
+"""
+macro TransformerWindingDispatch(
+    prefix,
+    postfix = "",
+    transformer_type = :ThreeWindingTransformer,
```
jd-lara, line 255: having to hardcore types here isn't great. I don't get why this approach using macros makes sense.

jd-lara (changes requested):
I am not convinced that using a macro for the three winding transformer here is necessary. It is way over engineered for a relatively simple function. Please provide more detail as to why this makes sense.

src/Ybus.jl
```diff
@@ -97,77 +97,104 @@ function add_to_branch_maps!(
...
-function _ybus!(
+function add_branch_entries_to_ybus!(
     y11::Vector{ComplexF32},
     y12::Vector{ComplexF32},
     y21::Vector{ComplexF32},
     y22::Vector{ComplexF32},
-    br::PSY.ACTransmission,
+    branch_ix::Int,
+    br::T,
+) where {T <: PSY.ACTransmission}
```
jd-lara, line 107: T isn't being used in this function

src/Ybus.jl
luke-kiernan: For this code, the only relevant difference between `Transformer3W` and the `PhaseShiftingTransformer3W` is whether the angle is called `phase_shift` or `\alpha`. I want to use multiple dispatch to define `_get_angle`, but for the other getters, one definition for the parent class of `ThreeWindingTransformer` suffices.

src/Ybus.jl
```diff
@@ -97,77 +97,104 @@ function add_to_branch_maps!(
...
+
+_get_angle(br::PSY.PhaseShiftingTransformer) = PSY.get_α(br)
+_get_angle(br::Union{PSY.TapTransformer, PSY.Transformer2W}) = PSY.get_phase_shift(br)
+
+_get_tap(::PSY.Transformer2W) = one(ComplexF32)
+_get_tap(br::Union{PSY.TapTransformer, PSY.PhaseShiftingTransformer}) = PSY.get_tap(br)
+
+"""Ybus entries for a `Transformer2W`, `TapTransformer`, or `PhaseShiftingTransformer`."""
+function ybus_branch_entries(
+    br::Union{PSY.Transformer2W, PSY.TapTransformer, PSY.PhaseShiftingTransformer},
```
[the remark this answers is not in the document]
luke-kiernan: Oh, I didn't realize there was a `TwoWindingTransformer`. Yeah, I'll do that.

luke-kiernan:
I'll remove the macros, and put the 3WT code back to how it was.

src/Ybus.jl
```diff
@@ -117,57 +129,113 @@ function _ybus!(
     adj[bus_to_no, bus_from_no] = -1
     fb[branch_ix] = bus_from_no
     tb[branch_ix] = bus_to_no
+    return
+end
+
+_get_shunt(br::PSY.ACTransmission, node::Symbol) =
```
jd-lara, line 135: There is some mild type piracy here but it isn't outrageous

jd-lara (approved):

PR #179, opened 2025-08-12
Mb/psi compat

m-bossart:

### DIFF
--- a/src/BA_ABA_matrices.jl
+++ b/src/BA_ABA_matrices.jl
@@ -53,16 +53,9 @@ end
 
 function BA_Matrix(ybus::Ybus)
     nr = ybus.network_reduction_data
-    direct_arcs = [x for x in keys(nr.direct_branch_map)]
-    parallel_arcs = [x for x in keys(nr.parallel_branch_map)]
-    series_arcs = [x for x in keys(nr.series_branch_map)]
-    transformer_arcs = [x for x in keys(nr.transformer3W_map)]
-    additional_arcs = [x for x in keys(nr.added_branch_map)]
     bus_ax = get_bus_axis(ybus)
     bus_lookup = get_bus_lookup(ybus)
-    arc_ax = unique(
-        vcat(direct_arcs, parallel_arcs, series_arcs, transformer_arcs, additional_arcs),
-    )
+    arc_ax = get_arc_axis(nr)
     n_isolated_buses = length(get_isolated_buses(ybus))
     n_entries = length(arc_ax) * 2 + n_isolated_buses
     BA_I = Vector{Int}(undef, n_entries)
--- a/src/Ybus.jl
+++ b/src/Ybus.jl
@@ -926,14 +926,7 @@ function _make_arc_subnetwork_axis(
     subnetworks::Dict{Int, Set{Int}},
     nr::NetworkReductionData,
 )
-    direct_arcs = [x for x in keys(nr.direct_branch_map)]
-    parallel_arcs = [x for x in keys(nr.parallel_branch_map)]
-    series_arcs = [x for x in keys(nr.series_branch_map)]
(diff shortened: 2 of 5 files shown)


src/BA_ABA_matrices.jl
```diff
@@ -94,6 +93,17 @@ function BA_Matrix(ybus::Ybus)
     return BA_Matrix(data, axes, lookup, subnetwork_axes, ybus.network_reduction_data)
 end
 
+function _get_series_susceptance(series_chain::Vector{Any})
+    series_susceptances = [_get_series_susceptance(segment) for segment in series_chain]
+    total_susceptance = 1 / (sum((1.0 ./ series_susceptances)))
+    return total_susceptance
+end
+function _get_series_susceptance(segment::PSY.Branch)
```
jd-lara, line 101: We need to tighten this type because it can dispatch on DC Branches. Use ACTransmission

src/BA_ABA_matrices.jl
```diff
@@ -94,6 +93,17 @@ function BA_Matrix(ybus::Ybus)
...
+    return PSY.get_series_susceptance(segment)
+end
+function _get_series_susceptance(segment::Set{PSY.Branch})
```
jd-lara, line 104: Same here

jd-lara (commented):
There are several instances that need a type review for the function dispatch from a generic branch to AC

src/BA_ABA_matrices.jl
```diff
@@ -71,10 +64,16 @@ function BA_Matrix(ybus::Ybus)
...
-        # TODO - should we consider phase shift?
-        b = 1 / imag(Zt)
+        # Get series susceptance from components, not the equivalent ybus for reductions of degree two nodes
+        # This results in reduced error relative to the DC power flow result without reductions
+        if haskey(ybus.network_reduction_data.series_branch_map, arc)
+            b = _get_series_susceptance(ybus.network_reduction_data.series_branch_map[arc])
+        else
+            Yt = -1 * ybus.data[ix_from_bus, ix_to_bus]
+            Zt = 1 / Yt
+            # TODO - should we consider phase shift?
```
jd-lara, line 75: record TODO in issue

src/Ybus.jl
```diff
@@ -842,93 +884,230 @@ function _make_subnetwork_axes(ybus, bus_numbers_to_remove, arcs_to_remove)
...
+        push!(I_ytf, row_ix)
+        push!(I_ytf, row_ix)
+        push!(J_ytf, equivalent_arc_indices[2])
+        push!(J_ytf, equivalent_arc_indices[1])
+        push!(V_ytf, ybus_boundary_isolated_d2_chain[2, 2])
+        push!(V_ytf, ybus_boundary_isolated_d2_chain[2, 1])
+        row_ix += 1
     end
+    yft_data = SparseArrays.sparse(I_yft, J_yft, V_yft, row_ix - 1, n_buses)
+    ytf_data = SparseArrays.sparse(I_ytf, J_ytf, V_ytf, row_ix - 1, n_buses)
```
jd-lara, line 958: Can you add a comment why you add a + 1 offset at the top loop and then remove it when making the sparse object?

src/common.jl
```diff
@@ -88,6 +88,11 @@ function get_arc_tuple(br::PSY.ACTransmission, nr::NetworkReductionData)
     get_arc_tuple(PSY.get_arc(br), nr)
 end
 
+#Assumes set of branches are in parallel 
+function get_arc_tuple(br::Set{PSY.ACTransmission}, nr::NetworkReductionData)
+    get_arc_tuple(PSY.get_arc(first(br)), nr)
+end
```
jd-lara, line 94: why it assumes that the set is in parallel? should we rename this function?

Issue #183, opened 2025-09-09
Increase test coverage for different branch types and reduction combinations


m-bossart:
Include tests that have double circuits and 3wt arcs in both radial and degree two reductions (4 combinations).
Requires new test systems or modification of existing systems.

PR #184, opened 2025-09-09
add tests where (parallel/3WT arc) is in a (series/radial) reduction.

luke-kiernan:
Address issue #183.


m-bossart:
> I'm aware that power flow in the loop is still broken in PSI: I'd like to have that fixed by mid-day friday. And since I'm the one who did the PF update for network reductions (which broke it), I feel that fixing it is my task.
> 
> The tests I wrote pass here in PNM: that's the important thing. I'll do the same checks in PF, make sure that the code works. After I have a PR in for fixing power flow in the loop, I will return to this and properly add these systems to PSB

I think this is a good plan, we don't need to merge this right now.

m-bossart (changes requested):
This just needs to be updated to use https://github.com/NREL-Sienna/PowerSystemsTestData/pull/105

m-bossart:
> Updated to use the PSB system, but now some tests fail. The degree 2 reduction is trying to reduce a loop, which messes with the logic of `_get_chain_data`. @m-bossart can you troubleshoot? If the series reduction shouldn't consider loops, then `degree_two_reduction.jl` needs to incorporate that somewhere...

Yes, I'll take a look this evening. We can add a check to avoid reducing a loop

m-bossart (commented):
I added the changes to avoid circular chains of 2d nodes. 
The tests pass

PR #187, opened 2025-09-15
Rb/add lcc hvdc

rbolgaryn:


jd-lara (commented):
@m-bossart we want to validate this against PSSe.

m-bossart:
> @m-bossart we want to validate this against PSSe.

I will work on adding a test with psse validation

m-bossart:
@rbolgaryn  What is the reason we need `data_facts` separate from the rest of the ybus?

PR #195, opened 2025-10-29
apply filters

jd-lara:
This PR enables the use of filters and better tracking to build optimization problems in PSI.

### DIFF
--- a/src/PowerNetworkMatrix.jl
+++ b/src/PowerNetworkMatrix.jl
@@ -320,7 +320,7 @@ get_lookup(mat::PowerNetworkMatrix) = mat.lookup
 
 function get_branch_multiplier(A::T, branch_name::String) where {T <: PowerNetworkMatrix}
     nr = A.network_reduction_data
-    if isempty!(nr.direct_branch_name_map)
+    if isempty(nr.direct_branch_name_map)
         populate_direct_branch_name_map!(nr)
     end
     if haskey(nr.direct_branch_name_map, branch_name)


src/NetworkReductionData.jl
```diff
@@ -7,22 +7,24 @@ network reduction algorithms.
...
 - `transformer3W_map::Dict{Tuple{Int, Int}, Tuple{PSY.ThreeWindingTransformer, Int}}`: Three-winding transformer mappings
 - `reverse_transformer3W_map::Dict{Tuple{PSY.ThreeWindingTransformer, Int}, Tuple{Int, Int}}`: Reverse transformer mappings
 - `removed_buses::Set{Int}`: Set of buses eliminated from the network
-- `removed_arcs::Set{Tuple{Int, Int}}`: Set of arcs eliminated from the network  
+- `removed_arcs::Set{Tuple{Int, Int}}`: Set of arcs eliminated from the network
 - `added_admittance_map::Dict{Int, Complex{Float32}}`: Admittances added to buses during reduction
 - `added_branch_map::Dict{Tuple{Int, Int}, Complex{Float32}}`: New branches created during reduction
 - `all_branch_maps_by_type::Dict{String, Any}`: Branch mappings organized by component type
 - `reductions::ReductionContainer`: Container tracking applied reduction algorithms
+- `name_to_arc_map::Dict{String, Tuple{Int, Int}}`: Maps string names to their corresponding arcs for modeling in optimization models or power flow reporting after reductions applied. It is possible to have repeated arcs for some names if case of serial or parallel combinations.
```
m-bossart:
```suggestion
- `name_to_arc_map::Dict{Type, Dict{String, Tuple{Tuple{Int, Int}, String}}}`: Maps string names to their corresponding arcs and the map where the arc can be found. Used in optimization models or power flow reporting after reductions are applied. It is possible to have repeated arcs for some names if case of serial or parallel combinations.
```

Issue #201, opened 2025-11-04
Handle parallel branch reductions for branches with phase shift


m-bossart:
For certain types of components `<: PSY.ACTransmission` we might want to not allow double circuit reductions (e.g. phase shifting transformers). We can add a check in `add_to_branch_maps!` when the `BranchesParallel` objects are created

m-bossart:
The current solution is to add a check and not add phase shifting transformers to `parallel_branch_map`:
https://github.com/NREL-Sienna/PowerNetworkMatrices.jl/blob/f29db913a65955b59b4ac8ff7e4163c0626e8d0f/src/Ybus.jl#L209-L214
This should be reexamined when developing modeling capabilities for phase shifting transformers.

PR #203, opened 2025-11-05
add additional methods

jd-lara:
These methods will enable proper support for PowerModels integration in PSI of the integration equivalents

### DIFF
--- a/src/definitions.jl
+++ b/src/definitions.jl
@@ -9,3 +9,8 @@ const ZERO_IMPEDANCE_LINE_REACTANCE_THRESHOLD = 1e-3
 const LODF_ENTRY_TOLERANCE = 1e-6
 
 DEFAULT_LODF_CHUNK_SIZE = 18_000
+
+SKIP_PARALLEL_REDUCTION_TYPES = [
+    PSY.PhaseShiftingTransformer,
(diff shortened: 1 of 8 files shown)


src/definitions.jl
```diff
@@ -9,3 +9,8 @@ const ZERO_IMPEDANCE_LINE_REACTANCE_THRESHOLD = 1e-3
 const LODF_ENTRY_TOLERANCE = 1e-6
 
 DEFAULT_LODF_CHUNK_SIZE = 18_000
+
+SKIP_PARALLEL_REDUCTION_TYPES = [
```
jd-lara: why do we have this? we can still calculate the values for a fixed alpha value.

m-bossart:
> I am generally ok with the changes but I don't understand why leave out the phase shifters.

The issue with the phase shifters is computing equivalent parameters when you have two branches in parallel with different phase angles. I don't think we can represent this case with a single set of branch parameters which is required for the powermodels translation:
https://github.com/NREL-Sienna/PowerNetworkMatrices.jl/blob/f29db913a65955b59b4ac8ff7e4163c0626e8d0f/src/common.jl#L323

m-bossart:
Related to [#201 ](https://github.com/NREL-Sienna/PowerNetworkMatrices.jl/issues/201)

PR #210, opened 2025-11-10
overload get_arc for 3wt winding

m-bossart:
Define `get_arc` for `ThreeWindingTransformerWinding`. This is hit in PSI when building the area interchanges


jd-lara:
a different fix is implemented in PSI.

PR #217, opened 2025-11-18
Review codebase for performance improvements

jd-lara:
- Optimize sparsify functions using droptol! instead of manual loops (10-50x speedup)
- Optimize Virtual LODF diagonal computation to avoid full O(n³) matrix inverse
- Optimize Ward reduction to use selective row solves instead of full matrix inversion
- Add LRU eviction tracking to RowCache for efficient cache management
- Pre-compute normalized distributed slack in VirtualPTDF constructor
- Add pre-allocated work arrays to VirtualPTDF and VirtualLODF to reduce allocations
- Use @inbounds for critical loops to improve performance
- Replace deepcopy with copy where appropriate to reduce memory overhead


jd-lara (commented):
@m-bossart I am ok with most of these changes. I tasted it and make sense, please take a look to approve and merge with the next release that include the other bug fixes for HVDC

src/ward_reduction.jl
jd-lara: @m-bossart there are some clever tricks here

PR #226, opened 2025-12-04
Lk/precompilation bugfix

luke-kiernan:
Make sure the extensions are actually tested and tell the CI to track code coverage for the extensions, too. [The CI changes were done by CoPilot: not sure how to test those/if the syntax is correct.]


test/runtests.jl
```diff
@@ -1,3 +1,11 @@
+# load system-specific packages before PNM: ensures relevant PNM extensions will be loaded.
+@static if (Sys.ARCH === :x86_64 || Sys.ARCH === :i686) && !Sys.isapple()
+    using MKL
```
jd-lara: Do you need to call using for both

jd-lara (approved):

PR #228, opened 2025-12-23
add emergency ratings

SebastianManriqueM:
This PR:

- Add the functions to get the equivalent emergency ratings for Series, Parallel and 3WTransformers.
- Fix the ``get_equivalent_rating()`` for ``BranchesSeries`` and Closes issue https://github.com/NREL-Sienna/PowerNetworkMatrices.jl/issues/234.
- This was tested togeter with my PSI branch for N-1: https://github.com/NREL-Sienna/PowerSimulations.jl/tree/sm/restore_n-1_post_psy5_clean_up
- It also creates new reduction mappings to handle reductions in N-1 models

# Notes
- the functions ``get_min_max_limits()`` remain PSI. That was my understanding after discussing with @m-bossart.
- I think we should discuss further Implementation of ``get_equivalent_rating()`` for ``BranchesParallel``, which currently is

```
function get_equivalent_rating(bp::BranchesParallel)
    # Sum of ratings divided by number of circuits
    return sum(PSY.get_rating(branch) for branch in bp.branches) / length(bp.branches)
end
```
My sugestion would be:
```
function get_equivalent_rating(bp::BranchesParallel)
    # Sum of ratings divided by number of circuits
    return sum(PSY.get_rating(branch) for branch in bp.branches) - maximum(PSY.get_rating(branch) for branch in bp.branches)
end
```

Notice that for emergency ratings this implementation considers the equivalent rating of ``BrancesParallel`` as the arithmetica addition of all the ratings.

### DIFF
--- a/src/BranchesSeries.jl
+++ b/src/BranchesSeries.jl
@@ -164,7 +164,7 @@ Return the emergency rating for PSY.ACTransmission branches.
 """
 function get_equivalent_emergency_rating(branch::PSY.ACTransmission)
     if isnothing(PSY.get_rating_b(branch))
-        @warn "Branch $(get_name(branch)) has no 'rating_b' defined. Post-contingency limit is going to be set using normal-operation rating.
+        @debug "Branch $(get_name(branch)) has no 'rating_b' defined. Post-contingency limit is going to be set using normal-operation rating.
             \n Consider including post-contingency limits using set_rating_b!()."
         return PSY.get_rating(branch)
     end


src/BranchesParallel.jl
```diff
@@ -76,6 +76,7 @@ function populate_equivalent_ybus!(bp::BranchesParallel)
     return
 end
 
+#TODO sm/further discuss this approach which will cause negligible capacity increasing when more than 2 circuits are in parallel
```
jd-lara: Do we need to resolve this TODO before merging?

src/BranchesSeries.jl
```diff
@@ -133,7 +133,45 @@ For series circuits, the rating is limited by the weakest link: Rating_total = m
...
+end
+
+"""
+    get_equivalent_emergency_rating(bs<:PSY.ACTransmission)
+
+Return the emergency rating for PSY.ACTransmission branches.
+Series chains, can be composed of PSY.ACTransmission branches and PNM.BranchesParallel.
+"""
+function get_equivalent_emergency_rating(branch::PSY.ACTransmission)
+    # Minimum rating for series branches (weakest link)
```
m-bossart: This comment is not relevant here

src/NetworkReductionData.jl
```diff
@@ -234,6 +249,12 @@ function populate_branch_maps_by_type!(nrd::NetworkReductionData, filters = Dict
                 Dict{PSY.ACTransmission, Tuple{Int, Int}}(),
             )
             map_by_type[k] = v
+            component_name_map = get!(
+                nrd.component_to_reduction_name_map,
+                _get_segment_type(k),
+                Dict{String, String}(),
+            )
+            #component_name_map[get_name(k)] = get_name(nrd.series_branch_map[v])
```
m-bossart: Need to add back the series_branch_map entries

src/BranchesParallel.jl
jd-lara: This is ok, we just need a reasonable documentation about the behavior.

src/BranchesSeries.jl
```diff
@@ -129,11 +129,46 @@ end
...
+end
+
+"""
+    get_equivalent_emergency_rating(bs<:PSY.ACTransmission)
+
+Return the emergency rating for PSY.ACTransmission branches.
+"""
+function get_equivalent_emergency_rating(branch::PSY.ACTransmission)
+    if isnothing(PSY.get_rating_b(branch))
+        @warn "Branch $(get_name(branch)) has no 'rating_b' defined. Post-contingency limit is going to be set using normal-operation rating.
```
jd-lara: Consider making this a debug.

daniel-thom:
```
@warn "Branch $(get_name(branch)) has no 'rating_b' defined. Post-contingency limit is going to be set using normal-operation rating." maxlog=50
```

src/NetworkReductionData.jl
```diff
@@ -63,6 +63,9 @@ network reduction algorithms.
         DataStructures.SortedDict{String, Tuple{Tuple{Int, Int}, String}},
     } =
         DataStructures.SortedDict{Type, Dict{String, Tuple{Tuple{Int, Int}, String}}}()
+    component_to_reduction_name_map::Dict{
+        Type,
+        DataStructures.SortedDict{String, String}} = DataStructures.SortedDict{Type, Dict{String, String}}()
```
m-bossart: Default is flipped. Inner Dict should be sorted

src/NetworkReductionData.jl
```diff
@@ -203,6 +212,12 @@ function populate_branch_maps_by_type!(nrd::NetworkReductionData, filters = Dict
                 Dict{BranchesParallel{_get_segment_type(k)}, Tuple{Int, Int}}(),
             )
             map_by_type[k] = v
+            component_name_map = get!(
+                nrd.component_to_reduction_name_map,
+                _get_segment_type(k),
+                Dict{String, String}(),
```
jd-lara: Same here

src/NetworkReductionData.jl
m-bossart: Remove sorted

jd-lara (approved):

PR #236, opened 2026-01-03
use repeated squaring in goderya

luke-kiernan:
For the goderya implementation, change from repeated multiplication by `T` (computing `T^i`, n steps) to repeated squaring (computing `T^(2^i)`, log(n) steps).

### DIFF
--- a/src/connectivity_checks.jl
+++ b/src/connectivity_checks.jl
@@ -1,28 +1,22 @@
 function _goderya(ybus::SparseArrays.SparseMatrixCSC)
     node_count = size(ybus)[1]
-    max_I = node_count^2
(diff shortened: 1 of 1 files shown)


luke-kiernan:
Two further improvements: we can use `rowval` (non-allocating) instead of `findnz` (allocating). Also, we can use `nnz` to figure out when the matrix stabilizes (so we're comparing numbers, not vectors).

luke-kiernan:
Seems like this goderya algorithm is legacy code. No point in optimizing it. Also, it's just slow: on a 10k bus system, union-find finishes in seconds, whereas goderya takes about a minute.

Issue #232, opened 2025-12-31
Radial reduction: diagonal entries in ybus


luke-kiernan:
I suspect the ybus isn't being computed correctly after a radial reduction: the `_apply_reduction` simply returns a slice of the old matrix. Example:
```julia

using PowerSystemCaseBuilder
using PowerNetworkMatrices
using PowerSystems

sys = build_system(MatpowerTestSystems, "matpower_case14_sys")

# Build Ybus with and without radial reduction
ybus_normal = Ybus(sys)
ybus_radial = Ybus(sys; network_reductions = NetworkReduction[RadialReduction()])
nrd = get_network_reduction_data(ybus_radial)
non_removed_bus = -1
for (k, v) in nrd.bus_reduction_map
    if length(v) > 0
        println("Buses ", v..., " reduced into bus ", k)
        non_removed_bus = k
    end
end
function in_removed_lines(br)
    (br.arc.from.number, br.arc.to.number) in nrd.removed_arcs ||
    (br.arc.to.number, br.arc.from.number) in nrd.removed_arcs
end
removed_line = only(get_components(br -> in_removed_lines(br), Line, sys))
i_radial = PowerNetworkMatrices.get_bus_lookup(ybus_radial)[non_removed_bus]
i_normal = PowerNetworkMatrices.get_bus_lookup(ybus_normal)[non_removed_bus]
println("impedance of removed line: ", removed_line.r + im * removed_line.x)
println("\nY[i,i] pre-radial reduction: ", ybus_normal[i_normal, i_normal])
println("Y[i,i] post-radial reduction: ", ybus_radial[i_radial, i_radial])
```
The impedance on the removed line is nonzero, yet the diagonal elements in the ybus matrix remain the same post-reduction.

PR #239, opened 2026-01-06
Fix ybus for radial reduction

m-bossart:
Modifies the Ybus after radial reduction to remove the contribution to diagonal entries from radial branches that have been removed. 

Closes [#232 ](https://github.com/NREL-Sienna/PowerNetworkMatrices.jl/issues/232)

### DIFF
--- a/test/test_ybus_reductions.jl
+++ b/test/test_ybus_reductions.jl
@@ -280,6 +280,7 @@ function set_radial_removed_arcs_to_unavailable!(sys, radial_removed_arcs, rbsm)
             end
         end
     end
+    return
 end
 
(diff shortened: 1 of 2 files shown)


test/test_ybus_reductions.jl
```diff
@@ -212,3 +212,95 @@ end
...
+                set_available!(l, false)
+                if arc.from.number ∈ keys(rbsm)
+                    set_available!(arc.from, false)
+                end
+                if arc.to.number ∈ keys(rbsm)
+                    set_available!(arc.to, false)
+                end
+            end
+        end
+    end
```
jd-lara: missing return

jd-lara (approved):
Add missing returns but this looks ok.

PR #243, opened 2026-01-12
suppress BLAS backend info message during precompilation

luke-kiernan:
As-is, if you have multiple dependencies that use PNM, then you get one message per dependency. This change makes it so you only get one, when you run `using PowerNetworkMatrices` for the first time.

### DIFF
--- a/src/PowerNetworkMatrices.jl
+++ b/src/PowerNetworkMatrices.jl
@@ -120,21 +120,17 @@ function _pardiso_single_LODF! end
 function _create_apple_accelerate_factorization end
 
 function __init__()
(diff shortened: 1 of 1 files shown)


luke-kiernan:
> Generating output whenever someone imports PowerSimulations without first importing an accelerator is still too noisy for my taste, given that this is the only message thus privileged.

I'll demote it to a debug message. We could just load the suggested BLAS backend, but I believe that deviates from typical Julia patterns. Packages usually leave BLAS backend as a runtime choice for the user and/or respect local environment variables, instead of forcing the user's hand.

PR #255, opened 2026-02-17
Add docstrings to all exported symbols

luke-kiernan:
## Summary
- Adds docstrings for the 2 exported symbols that were missing documentation
- `WardReduction`: struct for Ward equivalencing network reduction
- `get_network_reduction_data`: function to compute reduced network data

## Test plan
- [x] Docs build passes with no new errors
- [x] Formatter run clean

### DIFF
--- a/src/AdjacencyMatrix.jl
+++ b/src/AdjacencyMatrix.jl
@@ -244,7 +244,15 @@ function AdjacencyMatrix(ybus::Ybus)
 end
 
 """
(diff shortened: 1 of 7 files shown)


m-bossart (changes requested):
Shouldn't every method for an exported function have a docstring. For example, with `get_network_reduction_data` we have a method for each matrix, but only added a docstring to the method for `Ybus`

jd-lara (approved):

PR #270, opened 2026-03-18
virtual modf

jd-lara:
This a first approach to integration into PNM a VirtualMODF based on https://arxiv.org/pdf/2412.16164. There is a feature request to implement the calculation of the so called "FlowGates". 

This work will enable us to calculate the flow of a monitored element / contingency pair I.e., "what happens to line A if a specific outage of one or more elements happens".


src/virtual_lodf_calculations.jl
```diff
@@ -121,6 +136,63 @@ function _get_PTDF_A_diag(
...
+end
+
+"""
+    _extract_branch_susceptances_by_arc(BA, arc_ax, nr_data) -> Vector{Vector{Float64}}
+
+Extract per-branch susceptances for each arc. For arcs with a single branch,
+returns a one-element vector equal to the arc susceptance. For arcs with
+parallel branches (double circuits), returns one entry per branch.
+
+This enables single-branch contingencies on parallel arcs: when one circuit
```
m-bossart: Also enables single-branch contingencies on D2 reduced arcs.

PR #277, opened 2026-04-02
Ward fixes

m-bossart:
Fixes:
- Ward fails with singular value exception in klu when an HVDC connects to a single isolated bus.
- Bug in the handling of subnetwork axes when an entire subnetwork is eliminated via Ward. 

These fixes are needed for running Ward on WECC due to the representation of the hvdc interties to the EI.

### DIFF
--- a/src/ward_reduction.jl
+++ b/src/ward_reduction.jl
@@ -61,24 +61,23 @@ function get_ward_reduction(
         @error "no boundary buses found; cannot make bus_reduction_map based on impedance based criteria. mapping all external buses to the first reference bus ($first_ref_study_bus)"
         bus_reduction_map_index[first_ref_study_bus] = Set(external_buses)
     else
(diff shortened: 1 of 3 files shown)


src/ward_reduction.jl
```diff
@@ -100,7 +100,8 @@ function get_ward_reduction(
     end
 
     # Eq. (2.16) from  https://core.ac.uk/download/pdf/79564835.pdf
-    y_eq = y_be * KLU.solve!(klu(y_ee), Matrix{Complex{Float64}}(y_eb))
+    y_eq =
+        y_be * KLU.solve!(klu(y_ee; allowsingular = true), Matrix{Complex{Float64}}(y_eb))
```
jd-lara: This seems dangeous, can you elaborate on the change?

jd-lara (approved):

PR #280, opened 2026-04-03
add Ybus tracker for contingencies

### LINKED ISSUE #276: Create Delta Ybus object to track changes due to contingency.
we want to track effectively a sparse object that given a contingency stores the change in the  Ybus for efficient contingency calculations.

jd-lara:
closes #276

### DIFF
--- a/src/ybus_contingencies.jl
+++ b/src/ybus_contingencies.jl
@@ -431,25 +431,6 @@ function _process_outage_component!(
     return
 end
 
-# --- Arc lookup helper ---
-
-"""
-Find the arc tuple for a branch in the network reduction maps.
-Checks direct, parallel, and series maps in order.
-"""
-function _find_branch_arc(
-    nr::NetworkReductionData,
-    branch::PSY.ACTransmission,
-)
-    val = get(nr.reverse_direct_branch_map, branch, nothing)
-    val !== nothing && return val
-    val = get(nr.reverse_parallel_branch_map, branch, nothing)
-    val !== nothing && return val
-    val = get(nr.reverse_series_branch_map, branch, nothing)
-    val !== nothing && return val
-    error("Branch $(PSY.get_name(branch)) not found in any reduction map.")
-end
-
 # --- Constructors ---
 
 """
@@ -468,7 +449,6 @@ Uses a two-pass approach mirroring VirtualMODF's `_register_outage!`:
 
 # See Also
 - [`YbusModification(::Ybus, ::PSY.System, ::PSY.Contingency)`](@ref): Construct from contingency
-- [`YbusModification(::Ybus, ::PSY.ACTransmission, ::PSY.ACTransmission)`](@ref): Construct from impedance change
 """
 function YbusModification(
     ybus::Ybus,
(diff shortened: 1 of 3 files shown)


src/ybus_contingencies.jl
```diff
@@ -0,0 +1,631 @@
...
+# --- Combining modifications ---
+
+"""
+    $(TYPEDSIGNATURES)
+
+Combine two YbusModifications by adding their sparse deltas.
+Both modifications must originate from the same base Ybus.
+"""
+function Base.:+(m1::YbusModification, m2::YbusModification)
+    IS.@assert_op size(m1.data) == size(m2.data)
```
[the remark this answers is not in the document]
m-bossart: This is an interesting point. I wonder if we should store in each of the matrices the UUID of the system that they originate from? And then check that functions that take both a matrix and a system are consistent. For example: 
```
function YbusModification(
    ybus::Ybus,
    sys::PSY.System,
    contingency::PSY.Contingency,
)
```
This would allow us to fail with a meaningful error if the user does something invalid.

test/test_ybus_contingencies.jl
```diff
@@ -0,0 +1,359 @@
...
+        ACTransmission,
+        sys,
+    )
+        mod = YbusModification(ybus, PSY.Component[branch])
+        modified_data = apply_ybus_modification(ybus, mod)
+
+        # Reference: disable branch and rebuild
+        set_available!(branch, false)
+        ybus_ref = Ybus(sys)
+        set_available!(branch, true)
```
m-bossart: Not sure why we change the availability back to true in the tests. I believe this test should pass without this. This happens for a bunch of the following tests.

test/test_ybus_contingencies.jl
```diff
@@ -0,0 +1,359 @@
...
+    mod_ctg = YbusModification(ybus, sys, outage)
+
+    # Create modification from component directly
+    mod_direct = YbusModification(ybus, PSY.Component[line])
+
+    # Should be identical
+    @test isapprox(mod_ctg.data, mod_direct.data, atol = 1e-10)
+end
+
+@testset "YbusModification: PhaseShiftingTransformer rejection" begin
```
m-bossart: This test should use a test system that has phase shifting transformers.

test/test_ybus_contingencies.jl
```diff
@@ -0,0 +1,359 @@
...
+    # We need old_branch and new_branch to be the same object at different states.
+    # Since the line is mutated in-place, we compute the delta as:
+    #   ΔY = ybus_modified_ref.data - ybus_original.data
+    # and verify the constructor produces an equivalent result.
+    mod = YbusModification(ybus_original, line, line)
+
+    # Restore line to original state for the "old" entries
+    # Actually, the 3-arg constructor computes old and new from ybus_branch_entries
+    # on the same object. Since the line now has 2x, both old and new will be 2x,
+    # giving a zero delta. To properly test, restore and use two separate systems.
```
m-bossart: Similar comment here. The description ("restore and use two separate systems") does not match the code.

PR #281, opened 2026-04-03
performance large systems

jd-lara:


jd-lara:
Float32 values around ~2581 have absolute precision of ~±0.0002, and the series chain impedance change path compounds rounding through chain Ybus construction and subtraction. Windows math libraries round slightly differently, putting the error right at the 1e-4 boundary so I relaxed one of the tests.

Issue #294, opened 2026-04-23
Error in populate_branch_maps_by_type due to mixed parallel branch types.


m-bossart:
reported by @juflorez 
<img width="1423" height="711" alt="Image" src="https://github.com/user-attachments/assets/64969498-820e-4deb-ae8a-c631cc7a5c45" />

PR #295, opened 2026-04-24
remove parameterization of branches parallel

m-bossart:


m-bossart:
Fixes #294

m-bossart:
This is currently failing in the model build for WECC as well.

m-bossart:
This is not actually a parsing problem. We have been consistently parsing mixed parallel branch types from EI/WECC. The problem was introduced in https://github.com/Sienna-Platform/PowerNetworkMatrices.jl/pull/281.
My suggestion is to merge this PR and the accompanying PR in PSI (https://github.com/Sienna-Platform/PowerSimulations.jl/pull/1603) so that the released versions are compatible with the large datasets, and then to improve the performance of BranchesParallel once we have an established parsing strategy so we can guarantee not to have mixed type parallel branches.

PR #296, opened 2026-04-28
Add KLUWrapper

jd-lara:
This PR is important to reduce issues with KLU allowing for multi entry


src/KLUWrapper/klu_cache.jl
```diff
@@ -0,0 +1,281 @@
...
+"""
+function symbolic_factor!(cache::KLULinSolveCache{Tv},
+    A::SparseMatrixCSC{Tv, Int}) where {Tv}
+    n = _dim(cache)
+    if size(A, 1) != n || size(A, 2) != n
+        throw(DimensionMismatch(
+            "Cannot factor: cache is $(n)×$(n) but A is $(size(A)).",
+        ))
+    end
+    Base.finalize(cache)
```
luke-kiernan: You call `Base.finalize`, then proceed to modify the thing. A little sketchy.

luke-kiernan (commented):
One comment, will look more later.

src/KLUWrapper/solve_sparse_rhs.jl
```diff
@@ -0,0 +1,91 @@
...
+    Bcolptr = getcolptr(B)
+    Browval = rowvals(B)
+    Bnzval = nonzeros(B)
+
+    _ensure_scratch!(cache, block)
+    scratch = cache.scratch
+    col_map = cache.col_map
+
+    j_start = 1
+    @inbounds while j_start <= nb
```
luke-kiernan, line 50: Worth adding a comment here as to what this code is doing: expand columns of sparse RHS into dense columns in `scratch`, skipping columns that are structurally all zeros.

src/KLUWrapper/solve_sparse_rhs.jl
```diff
@@ -0,0 +1,91 @@
...
+        j_end = min(j_start + block - 1, nb)
+
+        npack = 0
+        for j in j_start:j_end
+            Bcolptr[j + 1] > Bcolptr[j] || continue
+            npack += 1
+            col_map[npack] = j
+            fill!(view(scratch, :, npack), zero(Tv))
+            for p in Bcolptr[j]:(Bcolptr[j + 1] - 1)
```
luke-kiernan: Nitpick: `Bcolptr[j]:(Bcolptr[j + 1] - 1)` is simply `nzrange(B, j)`.

jd-lara:
There is still something odd with the windows version that fails in a non deterministic way.

jd-lara:
I've looked at the copilots comments and addressed the one that look real

src/KLUWrapper/klu_cache.jl
```diff
@@ -0,0 +1,331 @@
...
+    end
+    if cache.symbolic != C_NULL
+        sym_ref = Ref(cache.symbolic)
+        klu_l_free_symbolic!(sym_ref, cache.common)
+        cache.symbolic = sym_ref[]
+    end
+    return nothing
+end
+
+# Public eager-release entry point. Overload of `Base.finalize` so callers can
```
luke-kiernan: Is this "separate public eager-release vs. private automatic GC release" distinction actually useful? Or is this just AI being overly fancy? Seems like the second to me.

src/KLUWrapper/pool.jl
```diff
@@ -0,0 +1,459 @@
...
+is left intact, mirroring the per-cache contract.
+"""
+function _free_klu_handles!(pool::KLULinSolvePool)
+    for cache in pool.workers
+        _free_klu_handles!(cache)
+    end
+    return nothing
+end
+
+# Public eager-release entry point at the pool level. Delegates to the
```
luke-kiernan: Same comment: is this really necessary?

src/solver_dispatch.jl
```diff
@@ -0,0 +1,108 @@
+# Shared dispatch for "run a solve under whatever solver we have." Lives
+# outside the Virtual{PTDF, LODF, MODF} files so all three matrices share
+# the same `with_solver` seam and the same KLU/AppleAccelerate factory.
+#
+# Windows libklu thread-safety: the MinGW build of `SuiteSparse_jll`'s
```
luke-kiernan: It can also error in non multi-threaded situations on windows, too, for seemingly inexplicable reasons. Do you have a minimum reproducible example of that? The only 2 examples I know of are both quite complex.

src/KLUWrapper/solve_sparse_rhs.jl
```diff
@@ -0,0 +1,91 @@
...
+`nrhs`. The cache reuses its packing buffer across calls; warm calls
+allocate nothing in the solver.
+
+Not thread-safe (mutates per-cache scratch). For parallel queries use
+`KLULinSolvePool`, where each worker owns an independent cache.
+"""
+function solve_sparse!(
+    cache::KLULinSolveCache{Tv},
+    B::SparseMatrixCSC{Tb, Int};
+    out::AbstractMatrix{Tv} = Matrix{Tv}(undef, _dim(cache), size(B, 2)),
```
luke-kiernan: Nitpick: if you're going to have a separate allocating version,`solve_sparse`, I'd just make `out` positional. Also kwargs sometimes have a performance cost (less compiler-friendly I believe).

src/row_cache.jl
```diff
@@ -176,3 +176,34 @@ function check_cache_size!(cache::RowCache; new_add::Bool = false)
...
+
+"""
+    cached_row_lookup(compute_row, cache, cache_lock, row, column, tol) -> value
+
+Shared cache-fast-path / compute / double-checked-insert pattern used by
+`VirtualPTDF._getindex` and `VirtualLODF._getindex`. Acquires `cache_lock`
+to test for a hit, runs `compute_row` outside the lock on a miss
+(KLU solves dominate the cost), then takes the lock again to insert. A
+concurrent producer that wins the insert race wins; the other side
+returns the winner's row. The signature places `compute_row` first so
```
luke-kiernan: > The signature places `compute_row` first so callers can use `do … end` block syntax.

explain?

src/solver_dispatch.jl
```diff
@@ -0,0 +1,108 @@
...
+- Anything else (e.g. `AppleAccelerate.AAFactorization`): serializes
+  through `solver_lock` and uses scratch slot `1`.
+
+`work_ba_col` and `temp_data` are `Vector{Vector{Float64}}` whose length
+matches the pool's `nworkers` (or `1` for the lock-serialized paths).
+"""
+function with_solver(
+    f::F,
+    K::KLULinSolvePool{Float64},
+    work_ba_col::Vector{Vector{Float64}},
```
luke-kiernan: Why did you go with a vector of vectors here, not a matrix?

src/KLUWrapper/pool.jl
```diff
@@ -0,0 +1,459 @@
...
+        end
+    end
+    return nothing
+end
+
+# Push one sentinel per worker into `pool.available`. Channel capacity is
+# `nworkers`, and the channel is empty when this is called (the caller has
+# just drained it), so all puts succeed immediately.
+function _push_dead_sentinels!(pool::KLULinSolvePool)
+    @lock pool.state_lock begin
```
luke-kiernan: These multi-threading things are confusing enough that I'd just check the invariant: `!isempty(pool.available) || throw("pool isn't empty")`

src/KLUWrapper/pool.jl
```diff
@@ -0,0 +1,459 @@
...
+# Drain every currently-valid worker from `available`, blocking on each
+# `take!` until the worker has been returned by its holder. Failed workers
+# (`pool.valid[i] == false`) are intentionally held out of the channel and
+# are already in the pool's exclusive ownership, so they are not part of
+# the drain count. Sentinels (`<= 0`) are not counted as valid; they live
+# in the channel only between a kill and the next `reset!`. The `pool.valid`
+# snapshot is taken under `state_lock` for consistency; the `take!` loop
+# blocks outside the lock so concurrent `release!` calls can land.
+function _drain_available!(pool::KLULinSolvePool)
+    n_to_drain = @lock pool.state_lock count(pool.valid)
```
luke-kiernan: Nitpick: there's a function for this, `n_valid(pool)`.

I'm also struggling to convince myself that there's no race condition here. Can a worker go from valid to invalid between the `count` call and the `take!` call? And if it can happen, does it cause issues?

src/KLUWrapper/pool.jl
```diff
@@ -0,0 +1,459 @@
...
+function reset!(pool::KLULinSolvePool{Tv},
+    A::SparseMatrixCSC{Tv, Int}) where {Tv}
+    @lock pool.admin_lock begin
+        n = nworkers(pool)
+        _drain_available!(pool)  # waits for in-flight valid workers
+        # If the pool was previously dead, sentinels may still be in the
+        # channel. Drain them so future `acquire!` calls consume valid
+        # indices instead.
+        _drain_sentinels!(pool)
+
```
luke-kiernan: After `_drain_sentinels!` and `drain_sentinels!`, should the pool be empty? I suspect so (in which case I'd add something to check said invariant)

luke-kiernan (changes requested):
I took a look at the backend multi-threading stuff--the matrix math itself I'm less familiar with.

src/solver_dispatch.jl
jd-lara: We are still investigating what's going on in Windows. The latest exploration showed that the issue only happens in windows when the PTDF and the LODF are used. Still not a good explanation of what's going on.

src/solver_dispatch.jl
jd-lara: _solve_factorization has three concretely-typed methods (KLU, generic, AppleAccelerate ext) all signed         
b::Vector{Float64}. Going to Matrix{Float64} + view gives a SubArray, forcing every signature to widen to StridedVector  which propagates into other areas of the code. I can add a comment about it

src/KLUWrapper/pool.jl
jd-lara: If you'd like to verify empirically, flip const `KLU_POOL_DEBUG = true` in KLUWrapper.jl and re-run the threaded testsets in test_virtual_threaded.jl; the per-worker atomic collision counter in with_worker               
(pool.jl:168-180) will log an @error if any task ever observes a worker held by another task concurrently.

src/KLUWrapper/pool.jl
jd-lara: I added more tests

src/row_cache.jl
jd-lara: explain what?

luke-kiernan: Why does placing `compute_row` first matter for `do...end` block syntax?

src/KLUWrapper/pool.jl
luke-kiernan: Hmm. Relying on empirical tests feels weird, because compiler optimizations and undefined behavior. But if you're confident this is ok, I'll take your word for it.

luke-kiernan (approved):
Looks good to me

jd-lara:
The final outcome of this PR is that we need a protection https://github.com/Sienna-Platform/PowerNetworkMatrices.jl/pull/296/commits/c71784329a76406db6ffb84245c734558b56ad0e to avoid the issue in KLU reported here https://github.com/DrTimothyAldenDavis/SuiteSparse/issues/1040 

After much testing, we can't parallelize the solves even with multiple pool workers and this at least resolves the issue with having a PTDF and MODF

src/virtual_ptdf_modification.jl
```diff
@@ -12,9 +12,9 @@ arcs, making this the recommended path for optimization loops where
 factors are computed once per modification and many rows are queried.
 
 !!! note
-    Parallel-safe on the KLU backend: scratch is per-worker and the solve
-    is routed through `KLULinSolvePool`. The AppleAccelerate backend
-    serializes through an internal lock — safe but not parallel.
+    Concurrent callers serialize on the per-cache `solver_lock` and
+    `_LIBKLU_LOCK` (KLU backend) or just the per-cache `solver_lock`
+    (AppleAccelerate backend).
```
luke-kiernan, line 17: Curious: did you try multi-threading with the AA backend?

luke-kiernan (approved):
Looks good.

src/KLUWrapper/KLUWrapper.jl
```diff
@@ -0,0 +1,88 @@
...
+access pattern KLU's user guide implies is supported). The corruption
+manifests two ways: an intermittent `KLU_INVALID` return with all input
+pointers still valid both pre- and post-call, and a `SIGSEGV` inside
+`klu_l_solve` (`klu_solve.c:118` in v7.8.3, the row-permutation read in
+the `nrhs == 1` chunk). The pre-call snapshot dump in `solve!` made
+both modes reproducible on macOS and confirmed the deterministic
+Windows-MinGW failure was the same bug.
+
+This lock is the only mechanism we have evidence for that prevents
+both modes; the `_solve_with_retry` recovery wrapper we briefly
```
josephmckinsey: This second comment about the _solve_with_retry should probably be removed.

src/KLUWrapper/klu_cache.jl
```diff
@@ -0,0 +1,334 @@
...
+    end
+    return nothing
+end
+
+# Public eager-release alias for `_free_klu_handles!` (the internal helper
+# stays unexported per the KLUWrapper convention).
+Base.finalize(cache::KLULinSolveCache) = _free_klu_handles!(cache)
+
+@inline function _check_pattern_match(cache::KLULinSolveCache,
+    A::SparseMatrixCSC, op::AbstractString)
```
josephmckinsey, line 206: Do we need to take `op` as a string given that the error call-stack will already include the calling function?

src/KLUWrapper/klu_jll_bindings.jl
josephmckinsey, line 1: Unlike KLU.jl, this doesn't look generated with Clang.jl (https://github.com/JuliaSparse/KLU.jl/blob/main/gen/generator.jl), so there is a bit of a risk this could get out of date. Why not use the same process as KLU.jl and put the klu_locks in the klu_cache.jl?

src/KLUWrapper/solve_sparse_rhs.jl
```diff
@@ -0,0 +1,107 @@
...
+            # in solve_dense.jl) to keep the inner-loop closure from capturing
+            # the per-chunk `npack`, which would force Julia to box it.
+            ok = _solve_call(
+                Tv, cache.symbolic, cache.numeric, n, Int64(npack),
+                pointer(scratch), cache.common,
+            )
+            if ok == 0 && cache.common[].status == KLU_INVALID
+                @warn "klu_solve (sparse RHS) returned KLU_INVALID; freeing numeric handle and re-factoring" cache_id =
+                    objectid(cache)
+                _recover_factorization!(cache)
```
josephmckinsey, line 80: I don't see a definition of `_recover_factorization!`.

src/KLUWrapper/solve_sparse_rhs.jl
```diff
@@ -0,0 +1,107 @@
...
+            npack += 1
+            col_map[npack] = j
+            fill!(view(scratch, :, npack), zero(Tv))
+            for p in rng
+                scratch[Browval[p], npack] = Bnzval[p]
+            end
+        end
+
+        if npack > 0
+            # Retry-on-KLU_INVALID is inlined here (mirrors `_solve_with_retry`
```
josephmckinsey: Another stray comment for the deleted `_solve_with_retry`.

src/BranchesParallel.jl
```diff
@@ -58,45 +80,89 @@ function compute_parallel_multiplier(
...
+Susceptance-weighted average of individual branch ratings,
+``\\sum_i f_i \\cdot S_i`` with ``f_i = b_i / \\sum_k b_k``. Reflects how DC flow
+physically splits across a parallel group. Throws `ArgumentError` if the total
+series susceptance is zero or non-finite.
 """
-function get_equivalent_rating(bp::BranchesParallel)
-    # Sum of ratings divided by number of circuits
-    return sum(get_equivalent_rating(branch) for branch in bp.branches) /
-           length(bp.branches)
+function get_impedance_averaged_rating(bp::AbstractBranchesParallel)
```
josephmckinsey, line 131: Probably not that important, but this appears to be O(n^2) where n is the number of branches in `bp`, and it creates a dict when it isn't necessary.

src/PowerflowMatrixTypes.jl
```diff
@@ -1,19 +1,20 @@
...
 }
-# VirtualPTDF can use various factorization types when extensions are loaded
+# VirtualPTDF can use various factorization types when extensions are loaded;
+# the K type parameter is left unconstrained on purpose.
 const DC_vPTDF_Matrix = VirtualPTDF{
     Tuple{Vector{Tuple{Int, Int}}, Vector{Int64}},
     Tuple{Dict{Tuple{Int, Int}, Int64}, Dict{Int64, Int64}},
-    <:LinearAlgebra.Factorization,
-}
+    K,
```
josephmckinsey, line 16: Any reason we can't make KLUinSolveCahce{Float64} a LinearAlgebra.Factorization like it is in KLU.jl?

src/ptdf_calculations.jl
```diff
@@ -137,33 +137,30 @@ function _calculate_PTDF_matrix_KLU(
...
+        "All buses are reference buses; PTDF is not defined.",
+    )
 
-    return
+    ABA = calculate_ABA_matrix(A, BA, ref_bus_positions)
+    cache = klu_factorize(ABA)
+    valid_ix = setdiff(1:buscount, ref_bus_positions)
+    PTDFm_t = zeros(buscount, linecount)
+    solve_sparse!(cache, BA[valid_ix, :], view(PTDFm_t, valid_ix, :))
+    PTDFm_t[collect(ref_bus_positions), :] .= 0.0
```
josephmckinsey: Very minor: If `valid_ix` is a setdiff with ref_bus_positions and `PTDFm_t` is initialized with 0s, then why do we need to set 0s again?

src/solver_dispatch.jl
```diff
@@ -0,0 +1,57 @@
...
+
+"""
+    _create_klu_solver(ABA) -> KLULinSolveCache{Float64}
+
+Factory used by `Virtual{PTDF, LODF, MODF}` constructors. Always returns
+a single `KLULinSolveCache`; concurrency is handled by the per-cache
+`solver_lock` at the wrapper level and by `_LIBKLU_LOCK` at the libklu
+boundary.
+"""
+function _create_klu_solver(ABA::SparseArrays.SparseMatrixCSC{Float64, Int})
```
josephmckinsey: This function isn't necessary at all anymore, is it? It just calls klu_factorization...

src/NetworkReductionData.jl
```diff
@@ -385,16 +390,27 @@ function populate_branch_maps_by_type!(nrd::NetworkReductionData, filters = Dict
...
+# A heterogeneous group must be discoverable under each of its members'
+# concrete branch types so that downstream per-type iteration (e.g. PSI's
+# `name_to_arc_map[MonitoredLine]`) can find it.
+_get_concrete_types(bp::MixedBranchesParallel) = unique(typeof.(bp.branches))
+
+# Construct an empty per-slot dict for `BranchMapsByType.parallel_branch_map`.
+# Value type is `AbstractBranchesParallel` so that the same per-type bucket can
+# hold either a homogeneous `BranchesParallel{T}` or a `MixedBranchesParallel`
+# that includes a branch of type `T`.
+_empty_parallel_branch_map(::AbstractBranchesParallel) =
```
luke-kiernan: Return value does not depend on argument--change to `_empty_parallel_branch_map()`?

src/BranchesParallel.jl
```diff
@@ -58,45 +80,89 @@ function compute_parallel_multiplier(
...
+end
+
+"""
+    get_single_element_contingency_rating(bp::AbstractBranchesParallel)
+
+N-1 rating for the parallel group: the surviving capacity after the largest-rated
+circuit trips, ``\\sum_i S_i - \\max_i S_i``. For a group of one branch this is zero.
+"""
+function get_single_element_contingency_rating(bp::AbstractBranchesParallel)
+    ratings = [get_equivalent_rating(branch) for branch in bp.branches]
```
luke-kiernan: nitpick: broadcastable. `ratings = get_equivalent_rating.(bp.branches)`. Similar for `get_sum_of_max_rating` too.

luke-kiernan (approved):
Two small nitpicks, but nothing major.

Will require companion PR in PF for compatibility.

test/test_network_modification.jl
```diff
@@ -250,3 +250,44 @@ end
...
+    # `src/virtual_ptdf_modification.jl`: many tasks calling
+    # `apply_woodbury_correction` on a single VirtualPTDF should agree with the
+    # serial baseline. The KLU path is exercised by the threaded testsets in
+    # `test/test_virtual_modf.jl`; this complements that coverage on the
+    # AppleAccelerate path.
+    if !PowerNetworkMatrices._has_apple_accelerate_ext()
+        @info "Skipping: AppleAccelerate extension not loaded."
+        return
+    end
+    if Threads.nthreads() < 2
```
josephmckinsey, line 265: As of Julia 1.12, Julia starts with 1 interactive thread and 1 default thread. It is unlikely this check is meaningful. In particular, I believe the github actions do not exercise multi-threading at all despite having no info about it.

josephmckinsey (changes requested):
Given how essentially every single call to KLU is globally serial, which is where all the run-time is, I'd be extremely surprised if a parallel call was ever faster. I'd definitely bet on it being thread-safe. I expect that we can relax _some_ things eventually with enough testing.

Most of these comments are probably not important, but the testing one and _recover_factorization are probably important.

src/KLUWrapper/klu_jll_bindings.jl
jd-lara: I wanted to tighten the scope to the types we use. I want to narrow the SuiteSparse_long entry points we actually call (klu_l_* / klu_zl_*); a Clang.jl pass would generate a bunch of code that needs post-filtering.

src/PowerflowMatrixTypes.jl
jd-lara: Do we need it?

src/KLUWrapper/klu_cache.jl
jd-lara: I wanted to make the error more readable when I was debugging, the stack trace was getting out of hand.

test/test_network_modification.jl
jd-lara: addressed in [c19b994](https://github.com/Sienna-Platform/PowerNetworkMatrices.jl/pull/296/commits/c19b994044d5725ac2ab00e111dad14614836c28)

src/BranchesParallel.jl
jd-lara: addressed in [c19b994](https://github.com/Sienna-Platform/PowerNetworkMatrices.jl/pull/296/commits/c19b994044d5725ac2ab00e111dad14614836c28)

src/KLUWrapper/solve_sparse_rhs.jl
jd-lara: fixed in [c19b994](https://github.com/Sienna-Platform/PowerNetworkMatrices.jl/pull/296/commits/c19b994044d5725ac2ab00e111dad14614836c28)

luke-kiernan (approved):
Changes since my last review look fine.

edit: Joseph's comments make me realize that just looking at "changes since last review" has its issues.

jd-lara:
> Changes since my last review look fine

Looking for the last tests from the PSI branch that consumes this

PR #302, opened 2026-05-14
aa wrapper and KLU migration to PNM for PFs

### LINKED ISSUE #107: Dependency on KLU.jl private interface
As of https://github.com/NREL-Sienna/PowerFlows.jl/pull/97, we are relying on significant private interface from [KLU.jl](https://github.com/JuliaSparse/KLU.jl). We should figure out a long-term plan for how to eliminate this unsafe dependency.

jd-lara:
This PR provides usage of the AA framework in the rest of the matrices and changes to KLU to be used later in PowerFlows to drop completely KLU.jl dependency and pass onto PFs the protection against the pointer problem and will close https://github.com/Sienna-Platform/PowerFlows.jl/issues/107


src/KLUWrapper/klu_cache.jl
```diff
@@ -53,72 +71,394 @@ is_factored(cache::KLULinSolveCache) =
...
+    sym_ref[] = reinterpret(Ptr{Cvoid}, typed[])
+    return nothing
+end
 
-@inline _refactor_call(::Type{Float64}, ap, ai, ax, sym, num, common) =
-    klu_l_refactor(ap, ai, ax, sym, num, common)
-@inline _refactor_call(::Type{ComplexF64}, ap, ai, ax, sym, num, common) =
-    klu_zl_refactor(ap, ai, ax, sym, num, common)
+# `klu_factor` — returns the numeric handle as a typed pointer; we
+# reinterpret to `Ptr{Cvoid}` for storage. Dispatch on both Tv and Ti.
```
jd-lara:
```suggestion
    colptr .-= one(Ti)
    rowval = Vector{Ti}(undef, length(rowvals(A)))
```

src/KLUWrapper/klu_cache.jl
```diff
@@ -53,72 +71,394 @@ is_factored(cache::KLULinSolveCache) =
...
+    sym,
+    num,
+    common,
+)
+    return klu_l_refactor(
+        ap, ai, ax,
+        reinterpret(SymbolicPtr, sym), reinterpret(NumericPtr, num),
+        common,
+    )
+end
```
jd-lara:
```suggestion
    cache::KLULinSolveCache{Tv, Ti},
) where {Tv, Ti}
```

src/KLUWrapper/klu_cache.jl
```diff
@@ -53,72 +71,394 @@ is_factored(cache::KLULinSolveCache) =
...
+    ap,
+    ai,
+    ax,
+    sym,
+    num,
+    common,
+)
+    return klu_z_refactor(
+        ap, ai, ax,
+        reinterpret(SymbolicPtr32, sym), reinterpret(NumericPtr32, num),
```
jd-lara:
```suggestion
        Tv, Ti,
        pointer(cache.colptr), pointer(cache.rowval),
```

src/KLUWrapper/klu_cache.jl
```diff
@@ -53,72 +71,394 @@ is_factored(cache::KLULinSolveCache) =
...
+    )
+end
+@inline function _solve_call(
+    ::Type{Float64},
+    ::Type{Int64},
+    sym,
+    num,
+    n,
+    nrhs,
+    b,
```
jd-lara:
```suggestion
`A`'s pattern, and run `klu_analyze` / `klu_l_analyze`.
```

src/KLUWrapper/klu_cache.jl
```diff
@@ -53,72 +71,394 @@ is_factored(cache::KLULinSolveCache) =
...
+    common,
+)
+    return klu_l_solve(
+        reinterpret(SymbolicPtr, sym), reinterpret(NumericPtr, num),
+        Int64(n), Int64(nrhs), b, common,
+    )
+end
+@inline function _solve_call(
```
jd-lara:
```suggestion
    if size(A, 1) != Int(n) || size(A, 2) != Int(n)
        throw(
            DimensionMismatch(
                "Cannot factor: cache is $(Int(n))×$(Int(n)) but A is $(size(A)).",
            ),
        )
```

src/KLUWrapper/klu_cache.jl
```diff
@@ -53,72 +71,394 @@ is_factored(cache::KLULinSolveCache) =
...
+    klu_l_free_numeric!(typed, common)
+    num_ref[] = reinterpret(Ptr{Cvoid}, typed[])
+    return nothing
+end
+@inline function _free_numeric!(
+    ::Type{ComplexF64},
+    ::Type{Int32},
+    num_ref::Ref{Ptr{Cvoid}},
+    common::Ref,
+)
```
jd-lara:
```suggestion
) where {Tv <: Union{Float64, ComplexF64}, Ti <: Union{Int32, Int64}}
    cache = KLULinSolveCache(
        A;
        reuse_symbolic = reuse_symbolic,
        check_pattern = check_pattern,
    )
```

src/AccelerateWrapper/aa_cache.jl
```diff
@@ -0,0 +1,348 @@
...
+release them eagerly.
+"""
+function AAFactorCache(
+    A::SparseMatrixCSC{Float64, Int};
+    reuse_symbolic::Bool = true,
+    check_pattern::Bool = true,
+    factorization_type::SparseFactorization_t = SparseFactorizationLDLT,
+)
+    n = size(A, 1)
+    n == size(A, 2) || throw(DimensionMismatch("matrix must be square; got $(size(A))"))
```
luke-kiernan, line 92: If the user tries to apply a symmetric factorization to a non symmetric matrix, this will silently take the lower triangle and factorize a different matrix...potentially hazardous. I'd at least check that the pattern is symmetric, with a flag to bypass.

src/AccelerateWrapper/aa_cache.jl
```diff
@@ -0,0 +1,348 @@
...
+end
+
+_pattern_mismatch(op::AbstractString) =
+    throw(ArgumentError("Cannot $op: matrix has different sparsity structure."))
+
+"""
+Release the libSparse numeric and symbolic handles held by `cache`, leaving
+Julia-side state intact. Idempotent.
+"""
+function _free_handles!(cache::AAFactorCache)
```
luke-kiernan, line 178: And if it isn't `SparseStatusOk`? Leaving it around seems iffy. Unlikely to cause issues in practice, but could clog up the heap with un-GCed objects in theory.

src/AccelerateWrapper/libsparse_bindings.jl
```diff
@@ -0,0 +1,326 @@
...
+# resolve it.
+function _libsparse_report_error(msg::Cstring)
+    s = unsafe_string(msg)
+    @error "libSparse reported an error" message = s
+    return nothing
+end
+
+# `reportError` is fired by libSparse before it returns a failure status —
+# we log the message so it ends up in user output rather than libSparse's
+# own stderr. Passing libc malloc/free explicitly (the C_NULL "use Apple
```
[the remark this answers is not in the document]
luke-kiernan: Seconded. Looking at `AppleAccelerate.jl`, I see instead: `@cfunction(text->error(unsafe_string(text)), Cvoid, (Cstring, ))`. Probably don't need to go that low-level, but `@error` seems risky.

src/KLUWrapper/klu_jll_bindings.jl
```diff
@@ -1,9 +1,32 @@
...
+# `KLU_INVALID` return; SEGV at `klu_solve.c:118`).
 
 import LinearAlgebra
 import SuiteSparse_jll: libklu
 
+# ---------------------------------------------------------------------------
+# Status codes and shared error helper
+# ---------------------------------------------------------------------------
+
+const KLU_OK = 0
```
luke-kiernan, line 20: nitpick: int backed enum would be more stylistic.

src/lodf_calculations.jl
```diff
@@ -384,7 +401,9 @@ efficient when the prerequisite matrices with factorization are already availabl
 
 # Keyword Arguments
 - `linear_solver::String = "KLU"`:
-        Linear solver algorithm for matrix computations. Currently only "KLU" is supported
+        This constructor is intentionally KLU-only because `ABA.K` is always a
```
luke-kiernan, line 404: Reasoning behind this decision? Which matrices are AA-compatible and why?

src/lodf_calculations.jl
```diff
@@ -64,6 +64,40 @@ get_network_reduction_data(M::LODF) = M.network_reduction_data
...
+#
+# The LODF computation builds a *diagonal* "demand" matrix `D = diag(m_V)`
+# where `m_V[i] = 1 - PTDF·A[i, i]` (clamped to 1.0 at `LODF_ENTRY_TOLERANCE`
+# to avoid divide-by-zero when an outage islands the line). The original
+# code factored `D` and ran a triangular solve `D · X = ptdf_denominator`;
+# that's a `factor + back-solve` over a diagonal, which collapses to
+# element-wise row scaling. KLU's BTF short-circuits this internally so the
+# overhead was modest; AA's libSparse and LAPACK's `getrf!`/`getrs!` do
+# not, so the previous code was 3–5× slower on AA and order-of-magnitude
+# slower on DENSE than necessary. Replace both with a direct row scaling.
```
luke-kiernan, line 77: Wait, is what we were doing that simple? And yet we were invoking factorization routines? Just want to confirm that this is real human error corrected by AI, not a hallucination

luke-kiernan (commented):
Big picture question: what's the motivation or goal here? Yeah, using the public Julia package has its issues and limitations, but it's also modular and maintained in tandem with the `_jll` binary. So I'd like to hear your reasoning here.

If there's specific shortcomings of these libraries, I'm willing to go open a PR. I expect it wouldn't get merged for a while (~months), but at least then we wouldn't need to maintain our own low-level bindings indefinitely.

My other question: why is AA no longer an extension?

jd-lara:
> Big picture question: what's the motivation or goal here? Yeah, using the public Julia package has its issues and limitations, but it's also modular and maintained in tandem with the `_jll` binary. So I'd like to hear your reasoning here.
> 
> If there's specific shortcomings of these libraries, I'm willing to go open a PR. I expect it wouldn't get merged for a while (~months), but at least then we wouldn't need to maintain our own low-level bindings indefinitely.
> 
> My other question: why is AA no longer an extension?

In short, I am trying reduce the number of external dependencies under which we have no control specially around the linear solvers and expose only the code that we need. I have been frustrated with the speed at which the other libraries respond and also at the exposure of mistakes there like it happened with the solve! call. 

Another reason specially for the linear solvers is to have cohesive infrastructure for PNM and PFs.

src/iterative_refinement.jl
josephmckinsey, line 1: What feature is this used for? I don't actually see any specific code for KLU, so it seems like it could potentially be used for any LinearAlgebra.Factorization.

src/AccelerateWrapper/AccelerateWrapper.jl
josephmckinsey, line 1: I'm a bit confused why we are wrapping this. KLU's internals weren't very thread-safe, which made it a clearer case until that can be fixed upstream, but I thought AppleAccelerate.jl was a bit better now?

src/KLUWrapper/klu_jll_bindings.jl
```diff
@@ -176,14 +321,305 @@ function klu_zl_free_numeric!(numeric_ref::Ref{NumericPtr},
...
 end
 
-# Status codes from klu.h.
-const KLU_OK = 0
-const KLU_SINGULAR = 1
-const KLU_OUT_OF_MEMORY = -2
-const KLU_INVALID = -3
-const KLU_TOO_LARGE = -4
+# ---------------------------------------------------------------------------
+# Int32 (int) family — mirror of the Int64 set above
```
josephmckinsey, line 325: Why are we implementing Int32? It seemed like the previous PR purposefully left out Int32

josephmckinsey (commented):
I'm mainly confused as to why we are making so many changes to KLU immediately again. I haven't dug into the Apple Accelerate part yet.

jd-lara:
> I'm mainly confused as to why we are making so many changes to KLU immediately again. I haven't dug into the Apple Accelerate part yet.

Maybe in not the cleanest way I layered the changes needed in KLU here so we can remove the use of KLU.jl in PowerFlows and use our own wrapper.  In PFs we use Int32

scripts/benchmarks/benchmark_solvers.jl
```diff
@@ -0,0 +1,376 @@
...
+    # ------------------------------------------------------------------
+    # VirtualPTDF: N_ROWS-row batch (fresh cold-cache object per pass)
+    # ------------------------------------------------------------------
+    if vptdf_ref !== nothing
+        try
+            println("  [VirtualPTDF] $N_ROWS-row batch …")
+            arc_keys = spread_keys(PNM.get_arc_axis(vptdf_ref), N_ROWS)
+            sample = function ()
+                v = PNM.VirtualPTDF(sys; linear_solver = solver)
+                GC.gc()
```
luke-kiernan, line 264: Why the explicit call to garbage collection here? Elsewhere in this file too.

src/AccelerateWrapper/aa_cache.jl
```diff
@@ -217,30 +211,35 @@ end
...
+    if new_nnz != cache.nnz
+        resize!(cache.rowIndices, new_nnz)
+        cache.nnz = new_nnz
     end
     if length(cache.columnStarts) != n + 1
         resize!(cache.columnStarts, n + 1)
     end
-    _populate_lower_triangle_pattern!(cache, A)
+    _populate_pattern!(cache, A)
     sym = _sparse_symbolic_factor(
```
luke-kiernan, line 232: nitpick: maybe rename this variable to something other than `sym` now that it's not a symmetric matrix.

src/AccelerateWrapper/solve_dense.jl
```diff
@@ -19,7 +19,10 @@ function solve!(cache::AAFactorCache, B::StridedMatrix{Cdouble})
     stride(B, 1) == 1 ||
         throw(ArgumentError("B must have unit stride in the first dimension."))
     size(B, 2) == 0 && return B
-    _sparse_solve_matrix!(cache.numeric, _dense_matrix(B))
+    # Workspace-aware overload — caller-supplied scratch avoids a per-call
+    # malloc/free inside libSparse.
+    ws = _ensure_solve_workspace!(cache, size(B, 2))
+    GC.@preserve cache _sparse_solve_matrix_ws!(cache.numeric, _dense_matrix(B), ws)
```
luke-kiernan, line 25: Is the `GC.@preserve` here precautionary or necessary?

src/iterative_refinement.jl
```diff
@@ -34,24 +53,47 @@ product plus one `solve!` against the cached factor.
...
-    size(B, 1) == Int(_dim(cache)) || throw(
+    size(B, 1) == _refine_dim(cache) || throw(
         DimensionMismatch(
-            "size(B, 1) = $(size(B, 1)); cache n = $(Int(_dim(cache))).",
+            "size(B, 1) = $(size(B, 1)); cache n = $(_refine_dim(cache)).",
         ),
     )
 
+    Tv = eltype(X)
     fill!(X, zero(Tv))
```
luke-kiernan, line 97: Nitpick: pretty sure just `fill!(X, 0.0)` will work. And if not, then introduce `Tv` as a type parameter.

src/linalg_settings.jl
```diff
@@ -14,25 +14,66 @@ _mkl_pardiso_install_error() =
     Install the Pardiso package:
     julia> using Pkg; Pkg.add(\"Pardiso\")"""
 
-_has_apple_accelerate_backend() = Sys.isapple()
+# Minimum macOS for the AppleAccelerate (libSparse LU) backend. The
+# `SparseFactorizationLU` code is API_AVAILABLE(macos(15.5)); older
+# libSparse rejects factorization type 80.
+const _AA_MIN_MACOS = v"15.5"
+
+# Query the running macOS product version via the `kern.osproductversion`
```
luke-kiernan, line 22: AppleAccerelate.jl has a very similar [function](https://github.com/JuliaLinearAlgebra/AppleAccelerate.jl/blob/master/src/AppleAccelerate.jl#L71). Could just copy-paste their implementation. [Or do you have reasons for not using theirs?]

src/linalg_settings.jl
```diff
@@ -14,25 +14,66 @@ _mkl_pardiso_install_error() =
...
-default for the `linear_solver` keyword on PTDF / LODF / VirtualPTDF /
-VirtualLODF / VirtualMODF constructors.
+Default sparse linear solver name. Returns "AppleAccelerateLU" on macOS
+$(_AA_MIN_MACOS.major).$(_AA_MIN_MACOS.minor)+ (Apple's built-in libSparse LU via `AccelerateWrapper`)
+and "KLU" elsewhere (non-Apple, or macOS older than $(_AA_MIN_MACOS.major).$(_AA_MIN_MACOS.minor)).
+Used as the default for the `linear_solver` keyword on PTDF / LODF /
+VirtualPTDF / VirtualLODF / VirtualMODF constructors.
 """
-_default_linear_solver() = Sys.isapple() ? "AppleAccelerate" : "KLU"
+function _default_linear_solver()
```
luke-kiernan, line 66: Nitpick: might belong better in `__init__` or in a similar compile-time "run just once" spot.

src/virtual_lodf_calculations.jl
```diff
@@ -114,6 +115,13 @@ function Base.show(io::IO, ::MIME{Symbol("text/plain")}, array::VirtualLODF)
     return
 end
 
+"""
+    _get_PTDF_A_diag(K, BA, A, ref_bus_positions) -> Vector{Float64}
+
+Compute `diag(PTDF · A)`. Each row of `A` has exactly two nonzeros (+1 at the
+from-bus, -1 at the to-bus), so the per-arc dot product reduces to two indexed
+reads into the solved PTDF row after a one-time transpose of `A`.
```
luke-kiernan, line 123: Pre-existing issue, but: is there really no better way to do this? I have a sense of deja vu, as if I've gone down this path before and it went nowhere.

edit: okay this has lower memory usage than pre-computing the entirety of `K^{-1}` (which could be dense). I'm guessing that's why we pick this method.

luke-kiernan (approved):

Issue #301, opened 2026-05-13
Removing islands causes issue with radial reductions


acostarelli:
In the Argentina system if we attempt to remove the islands in some form like so:

```julia
set_available!.(get_components(x -> !get_available(x), ACBranch, sys), true)

for g in get_components(x -> get_bustype(x) == ACBusTypes.ISOLATED, ACBus, sys)
    set_bustype!(g, ACBusTypes.PQ)
end

for g in get_components(ACBus, sys)
    set_available!(g, true)
end
```

There is an error later when trying to do radial reductions:
```julia
Ybus(sys, network_reductions=NetworkReduction[RadialReduction()])
```

Inspecting the subnetwork matrices of the Ybus without network reductions shows a lot of isolated buses. Something to do with 3W transformers, which are marked available but their windings are not -- and setting their windings to available doesn't fix this, nor does in addition skipping the step of setting Isolated buses to PQ.

@SebastianManriqueM 
@m-bossart

m-bossart:
According to the debugging by @acostarelli the root cause of this error is a closed breaker/switch which meets the zero impedance line reduction criteria in parallely with a line that is set to available. The fix for this specific case is probably to not make those lines available or to re-examine the modification of component availability. 

However, this case should still be able to run without error. This requires a refactoring of the zero impedance line reduction. This refactoring is in the works already to address #299

Issue #299, opened 2026-05-12
Lines with near-zero impedance


luke-kiernan:
We're only contracting `DiscreteControlledACBranch` components, not other lines with very small `r` and `x`. Is this a purposeful design choice?

It's causing loss of precision when running power flows: I'm seeing `|y[i,j]| ~ 1e4` with a voltage difference of about `1e-4` across the line, so then I get a discrepancy of about `Y * ΔV ~ 1e0` in the KCL calculation. And so I get a non-zero residual at the bus, on the order of mangitude of a couple p.u.

edit: I'm not sure the above accounting is quite right, but basically...small `x` and `r` -> huge `y` -> loss of precision, from `1e5 - 1e5 = 0±1e-2`, since `Float32` only has 7 decimal places of precision.

I guess another fix would be to allow for the user to specify the ZIL threshold.

jd-lara:
@m-bossart what do you think? This seems like more data challenges in general. I event think this probably should go on the parser.

m-bossart:
I think the parsing is probably fine. I do think PSSE considers all non-transformer branches for the zero impedance check and we should probably do the same. I also think the x threshould should be changed from 1e-3 to 1e-4 to match the PSSE default. @mcllerena do you remember why we went with 1e-3?

mcllerena:
Not sure why it is set to that here, in PSY is set originally to 1e-4
https://github.com/Sienna-Platform/PowerSystems.jl/blob/main/src/definitions.jl#L580

PR #307, opened 2026-05-20
handle zero impedance branches within reductions structure

m-bossart:
Addresses #299 and #301


src/Ybus.jl
```diff
@@ -433,7 +433,13 @@ _get_shunt(::PSY.DiscreteControlledACBranch, ::Symbol) = zero(YBUS_ELTYPE)
 
 """Ybus entries for a `Line` or a `DiscreteControlledACBranch`."""
 function ybus_branch_entries(br::PSY.ACTransmission)
-    Y_l = (1 / (PSY.get_r(br) + PSY.get_x(br) * 1im))
+    r = PSY.get_r(br)
```
jd-lara, line 436: Shouldn't we just eliminate the branch ? According to PSSe doc this is equivalent to the switch and the two nodes are the same

m-bossart: If you have a branch with shunt components we want to retain those contributions to the ybus. The new strategy implemented in this PR is to add all of the branches (even the ones with close to zero impedance) and then update the ybus by "merging" those two nodes based on the off-diagonal ybus entries. So in effect those two nodes are the same, it just happens separate from the ybus build (like the other reductions)

m-bossart:
@jd-lara There are some comments from co-pilot that I need to address, including better testing. But I'm interested in feedback on the overall approach of just adding all of the components and then looking at the Ybus to determine which branches qualify for the zero-impedance reduction. The benefit of this is not having to worry about wacky combinations of parallel devices in the data causing issues.

src/Ybus.jl
```diff
@@ -1365,6 +1439,65 @@ function _apply_bus_reductions!(nr::NetworkReductionData, nr_new::NetworkReducti
     return bus_numbers_to_remove
 end
 
+function _remap_arc_keys_batch!(
```
jd-lara: There should be at least a @debug call somewhere here to avoid making a silent overwrite

src/zero_impedance_branch_reduction.jl
```diff
@@ -0,0 +1,74 @@
...
+        end
+    end
+    return nothing
+end
+
+function _build_transformer_arc_set(nrd::NetworkReductionData)
+    transformer_arcs = Set{Tuple{Int, Int}}()
+    union!(transformer_arcs, keys(nrd.transformer3W_map))
+    for (arc, branch) in nrd.direct_branch_map
+        branch isa PSY.TwoWindingTransformer && push!(transformer_arcs, arc)
```
jd-lara: Use dispatch here not a isa

src/definitions.jl
```diff
@@ -6,6 +6,8 @@ const GiB = MiB * KiB
 const MAX_CACHE_SIZE_MiB = 100
 const ROW_PERSISTENT_CACHE_WARN = 1 * GiB
 const ZERO_IMPEDANCE_LINE_REACTANCE_THRESHOLD = 1e-3
```
jd-lara: This is unused now. Remove if its no longer applied elsewhere in the code

src/zero_impedance_branch_reduction.jl
```diff
@@ -0,0 +1,74 @@
...
+    reduction::ZeroImpedanceBranchReduction,
+)
+    nr = NetworkReductionData()
+    nrd = get_network_reduction_data(ybus)
+    bus_ax = get_bus_axis(ybus)
+    transformer_arcs = _build_transformer_arc_set(nrd)
+    nz_rows, nz_cols, nz_vals = SparseArrays.findnz(get_data(ybus))
+    for (row_ix, col_ix, v) in zip(nz_rows, nz_cols, nz_vals)
+        row_ix >= col_ix && continue
+        real(v) == 0.0 || continue
```
jd-lara: This entries will be Float32 so you will want to use a ` iszero(real(v))` to avoid memory costs on conversion. 

This might apply in other places

src/Ybus.jl
```diff
@@ -1276,19 +1295,70 @@ function _resolve_arc_admittance(
...
+    bus_lookup::Dict{Int, Int},
+    merged_bus_pairs::Dict{Int, Int},
+)
+    for (removed_bus, surviving_bus) in merged_bus_pairs
+        i = bus_lookup[surviving_bus]
+        j = bus_lookup[removed_bus]
+        data[i, :] += data[j, :]
+        data[:, i] += data[:, j]
+        adjacency_data[i, :] += adjacency_data[j, :]
+        adjacency_data[:, i] += adjacency_data[:, j]
```
jd-lara: All of this block is allocating. Each data[i, :] += something has 2  SparseVector materialization and triggers a full traverseasl of the `colptr` in the sparse matrix. 

You need dedicated accumulators. 

If a system has 1000s of zero impedance objects this loop becomes costly

src/Ybus.jl
jd-lara, line 1225: Needs to be type signed

jd-lara (changes requested):
The biggest comments is to avoid == 0.0 and use sizer and the refactoring of the accumulators. Those can be re-written as 

```Julia
function _accumulate_csc_row_into!(
      M::SparseArrays.SparseMatrixCSC,
      i::Int,
      j::Int,
  )
      rows = SparseArrays.rowvals(M)
      vals = SparseArrays.nonzeros(M)
      for col in 1:size(M, 2)
          v_j = zero(eltype(M))
          for k in SparseArrays.nzrange(M, col)
              rows[k] == j && (v_j = vals[k])
          end
          iszero(v_j) && continue
          found = false
          for k in SparseArrays.nzrange(M, col)
              if rows[k] == i
                  vals[k] += v_j
                  found = true
                  break
              end
          end
          found || (M[i, col] += v_j)
      end
      return
  end

  function _accumulate_csc_col_into!(
      M::SparseArrays.SparseMatrixCSC,
      i::Int,
      j::Int,
  )
      rows = SparseArrays.rowvals(M)
      vals = SparseArrays.nonzeros(M)
      j_range = SparseArrays.nzrange(M, j)
      i_range = SparseArrays.nzrange(M, i)
      for k_j in j_range
          r = rows[k_j]
          v_j = vals[k_j]
          iszero(v_j) && continue
          found = false
          for k_i in i_range
              if rows[k_i] == r
                  vals[k_i] += v_j
                  found = true
                  break
              end
          end
          found || (M[r, i] += v_j)
      end
      return
  end

  function _merge_ybus_buses!(
      data::SparseArrays.SparseMatrixCSC{YBUS_ELTYPE, Int},
      adjacency_data::SparseArrays.SparseMatrixCSC{Int8, Int},
      bus_lookup::Dict{Int, Int},
      merged_bus_pairs::Dict{Int, Int},
  )
      for (removed_bus, surviving_bus) in merged_bus_pairs
          i = bus_lookup[surviving_bus]
          j = bus_lookup[removed_bus]
          _accumulate_csc_row_into!(data, i, j)
          _accumulate_csc_col_into!(data, i, j)
          _accumulate_csc_row_into!(adjacency_data, i, j)
          _accumulate_csc_col_into!(adjacency_data, i, j)
      end
      return
  end
```

jd-lara (approved):

PR #309, opened 2026-05-27
add monitored components to exceptions automatically

jd-lara:
This PR makes sure that if reductions are applied then the outages monitored components buses are automatically added to the exceptions list to prevent incorrect ABA calculations


jd-lara:
@m-bossart this PR surfaced a bug in PFs reporting code fixed here https://github.com/Sienna-Platform/PowerFlows.jl/pull/377/commits/3d7a71c08b590613c31a06b62cf998a7c296c9d4

src/virtual_modf_calculations.jl
```diff
@@ -65,7 +65,10 @@ cache and skips the recomputation.
 - `max_cache_size_bytes::Int`:
         Max cache size in bytes per contingency.
 - `network_reduction_data::NetworkReductionData`:
-        Network reduction mappings for branch resolution.
+        Network reduction mappings for branch resolution. The retained buses are the
+        keys of its `bus_reduction_map`; its `irreducible_buses` set is the narrower
```
m-bossart: This docstring makes it seem like `irreducible_buses` is a filed of the NetworkReductionData which it is not

src/virtual_modf_calculations.jl
```diff
@@ -274,13 +283,28 @@ function VirtualMODF(
...
+    if !isempty(dist_slack)
         @info "Distributed bus"
     end
     solver = resolve_linear_solver(linear_solver)
 
+    # Mandatory: the ABA/Woodbury solve runs on the reduced network, so a branch in
+    # a contingency must survive reduction or the matrices won't match the queries.
+    # Protected buses land in the resulting `NetworkReductionData` (no copy kept).
+    if isempty(network_reductions)
+        applied_reductions = network_reductions
```
m-bossart: If `network_reductions` from the user is empty, we still will run the zero-impedance reduction. Don't we still need to protect buses to make sure they aren't reduced there?

src/virtual_modf_calculations.jl
```diff
@@ -274,13 +283,28 @@ function VirtualMODF(
...
+    else
+        protected_buses = _collect_protected_buses(sys)
```
m-bossart: This function builds the full Ybus from the system. So we end up building the Ybus twice.

src/Ybus.jl
```diff
@@ -1219,6 +1268,7 @@ function build_reduced_ybus(
         get_reductions(get_network_reduction_data(ybus)),
     )
     network_reduction_data = get_reduction(ybus, sys, network_reduction)
+    isempty(network_reduction_data) && return ybus
```
m-bossart, line 1302: This seems unnecessary. In what case would we hit this?

src/Ybus.jl
```diff
@@ -1367,18 +1417,12 @@ function _accumulate_csc_col_into!(
...
-    j_len = length(j_range_init)
-    # Structural inserts into column i (which precedes j in CSC order) shift colptr[j]
-    # upward, invalidating any pre-captured range. Track cumulative shift in `offset` so
-    # the absolute position of entry idx in column j's backing store remains j_start + offset + idx.
-    offset = 0
-    for idx in 0:(j_len - 1)
-        k_j = j_start + offset + idx
-        r = rows[k_j]
-        v_j = vals[k_j]
+    # Snapshot column j up front. Structural inserts into column i shift the CSC backing
```
m-bossart, line 1451: I'm not sure I fully understand the tradeoff for these two implementations.

src/AdjacencyMatrix.jl
```diff
@@ -128,67 +128,86 @@ end
...
-    sys::PSY.System,
-    reduction::DegreeTwoReduction,
-)
-    irreducible_buses = get_irreducible_buses(reduction)
-    validate_buses(A, irreducible_buses)
-    network_reduction_data = get_network_reduction_data(A)
-    direct_branch_map = network_reduction_data.direct_branch_map
-    parallel_branch_map = network_reduction_data.parallel_branch_map
-    transformer3W_map = network_reduction_data.transformer3W_map
+# Push the available endpoint bus numbers of a branch into `buses`. Set-typed so
```
m-bossart: Why is this code in AdjacencyMatrix.jl

src/AdjacencyMatrix.jl
```diff
@@ -128,67 +128,86 @@ end
...
+    _add_arc_buses!(buses, PSY.get_arc(br))
+
+function _add_arc_buses!(buses::Set{Int}, br::PSY.ThreeWindingTransformer)
+    _add_arc_buses!(buses, PSY.get_primary_star_arc(br))
+    _add_arc_buses!(buses, PSY.get_secondary_star_arc(br))
+    _add_arc_buses!(buses, PSY.get_tertiary_star_arc(br))
+    return
+end
+
+"""
```
m-bossart: Same as above here. Seems like a lot of generic code that belongs somewhere else.

src/ReductionContainer.jl
```diff
@@ -1,39 +1,90 @@
...
+end
+
+# ZIBR is auto-applied by `Ybus(sys; ...)`; a user-supplied one would double-apply.
+_reject_zibr_in_user_reductions(::NetworkReduction) = nothing
+function _reject_zibr_in_user_reductions(::ZeroImpedanceBranchReduction)
+    throw(
+        IS.DataFormatError(
+            "ZeroImpedanceBranchReduction is auto-applied during Ybus construction; " *
+            "do not include it in `network_reductions`. To customize its detection " *
+            "threshold or substituted reactance, pass `zero_impedance_reduction = " *
```
m-bossart, line 37: It doesn't seem like we test anywhere the customization path for the zero impedance reduction

src/virtual_modf_calculations.jl
```diff
@@ -271,16 +280,24 @@ function VirtualMODF(
...
         @info "Distributed bus"
     end
     solver = resolve_linear_solver(linear_solver)
 
-    # Build network matrices (same path as VirtualLODF)
-    Ymatrix = Ybus(sys; network_reductions = network_reductions, kwargs...)
+    # Outage/monitored buses are auto-protected so contingency branches survive
+    # reduction. Registering an outage with previously-unseen monitored components
+    # after construction shifts this set and requires rebuilding the MODF.
+    protected_buses = _collect_protected_buses(sys, irreducible_buses)
```
m-bossart: _collect_protected_buses builds the full Ybus to find the protected buses. It still seems like we have unnecessary repeated work with this design.

src/ReductionContainer.jl
jd-lara: This is added now

src/Ybus.jl
jd-lara: The old version tracked CSC backing-store offsets assuming inserts into column i shift the storage in a fixed direction this is only true when i < j. Once we added the zero-impedance merges where the survivor index i > removed index j, that bookkeeping dropped a mutual term, producing an asymmetric Ybus. This is where the PowerFlow tests started to fail

src/Ybus.jl
jd-lara: just a short circuit to avoid entering the other reduction analysis

test/test_modf_reduction_consistency.jl
```diff
@@ -0,0 +1,404 @@
...
+        outaged = br
+        break
+    end
+    @test outaged !== nothing
+
+    PSY.add_supplemental_attribute!(
+        sys,
+        outaged,
+        _fixed_outage(; monitored = [monitored]),
+    )
```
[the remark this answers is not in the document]
jd-lara: all of these are false positive

PR #310, opened 2026-05-27
simplify _update_bus_maps!

m-bossart:


jd-lara:
@m-bossart I incorporated this un #309 with testing.

PR #311, opened 2026-05-31
use the helpers to reduce reactive power

### LINKED ISSUE #222: Fix the reduce reactive power injectors flag
The reduce reactive power injectors flag exists but doesn't currently do anything.

jd-lara:
Improving the reduction code to enable the reduction injectors that only supply reactive power. 

This PR requires https://github.com/Sienna-Platform/PowerSystems.jl/pull/1645

Fixes #222


m-bossart (approved):
This looks good assuming the tests pass with the PSY update with the new traits

PR #317, opened 2026-06-02
handle reverse parallel arc

m-bossart:


jd-lara:
### Code review

Found 1 issue:

1. Anti-parallel branches are merged into one `BranchesParallel` group keyed by `reverse_new_arc`, but the group's orientation invariant is not preserved. The incoming branch is stored under the reverse direction without flipping its admittance block. This is correct for the DC path (BA/ABA/PTDF/LODF use the orientation-independent scalar `get_series_susceptance`, which is why the fix works there), but the AC equivalent path sums each member's 2x2 Y-block positionally and explicitly assumes all members share orientation:

https://github.com/Sienna-Platform/PowerNetworkMatrices.jl/blob/59faae9cd59a89589ce3ad6b5a9bc01107a26b5d/src/Ybus.jl#L471-L486

The new collision cases violate that invariant by placing a branch oriented `new_arc = (A,B)` into a group keyed `reverse_new_arc = (B,A)`:

https://github.com/Sienna-Platform/PowerNetworkMatrices.jl/blob/59faae9cd59a89589ce3ad6b5a9bc01107a26b5d/src/Ybus.jl#L1631-L1645

(same pattern in the parallel→reverse cases at L1676-L1685). For symmetric lines `Y11=Y22` and `Y12=Y21`, so the positional sum is unaffected. For asymmetric branches (transformers, phase shifters), `Y11≠Y22` and `Y12≠Y21`, so the reverse-oriented member's tap/shunt admittance lands on the wrong port and `get_equivalent_physical_branch_parameters` returns an incorrect equivalent Ybus. Consider swapping the reverse member's `(Y11,Y12,Y21,Y22)` to `(Y22,Y21,Y12,Y11)` before grouping, or normalizing orientation, so the L478 invariant continues to hold.

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

jd-lara (approved):

PR #319, opened 2026-06-05
tolerance improvement

jd-lara:
This PR is adding an auto detection of the sparsification for the NetworkMatrices based on the properties of the Ybus and de desired precision. 

The concept is to find a alpha "safety fraction". This keeps the sparsity decision scale-invariant: each row is thresholded against its own largest coefficient, so a row whose entries are all O(10^{-3}) is not wiped out, and a row with O(1) entries isn't left fully dense.

In the EI case this approach automatically reduced 50% of the entries automatically, I need to test the loss vs the DC PF, this is why I am leaving as a draft.


src/auto_tolerance.jl
```diff
@@ -0,0 +1,270 @@
...
+# rtol. Returns maxs when no small s fits (full precision), 0 for zero/non-finite.
+function _sigfigs(v::Float64, maxs::Int, rtol::Float64)
+    (iszero(v) || !isfinite(v)) && return 0
+    av = abs(v)
+    e = floor(Int, log10(av))                 # decade of the leading digit
+    for s in 1:maxs
+        factor = 10.0^(s - 1 - e)
+        vr = round(av * factor) / factor
+        abs(vr - av) <= rtol * av && return s
+    end
```
luke-kiernan, line 98: According to Claude, equivalent to:
```julia
for s in 1:maxs
    abs(round(av; sigdigits = s) - av) <= rtol * av && return s
end
```

src/auto_tolerance.jl
```diff
@@ -0,0 +1,270 @@
...
+    return maxs
+end
+
+# Linear-interpolated quantile of an already-sorted integer vector. Avoids a
+# Statistics dependency for the one place we need a quantile.
```
luke-kiernan: > Avoids a Statistics dependency

Statistics is a julia stdlib, not an external dependency.

src/auto_tolerance.jl
```diff
@@ -0,0 +1,270 @@
...
+        spec.safety * delta,
+        MIN_RELATIVE_TOLERANCE,
+        MAX_RELATIVE_TOLERANCE,
+    )
+end
+
+# Hager's algorithm for ‖B‖₁ where `applyB!(v)` overwrites v with B·v. B = ABA⁻¹
+# is symmetric here, so Bᵀ·v = B·v and only the forward solve is needed. Bounded
+# iterations guard against non-convergence; γ is the running lower bound.
+function _hager_invnorm1(applyB!, n::Int; itmax::Int = 5)
```
luke-kiernan, line 169: Only used in a logging message. Claude says there's better versions of the algorithm: this one "lacks the standard safeguard against re-selecting a previous unit vector."

Aside: apparently LAPACK's dlacn2 would be the drop-in here, but there's no Julia wrapper...I could write one if desired.

src/auto_tolerance.jl
luke-kiernan: Also, non-power-of-10 base powers would mess with this logic: for such systems, converting between units (system base vs. natural units) causes artificial gain/loss of precision.

src/auto_tolerance.jl
```diff
@@ -0,0 +1,270 @@
...
+    n = length(v)
+    n == 1 && return Float64(v[1])
+    h = (n - 1) * q + 1.0
+    lo = floor(Int, h)
+    lo >= n && return Float64(v[n])
+    frac = h - lo
+    return v[lo] + frac * (v[lo + 1] - v[lo])
+end
+
+"""
```
luke-kiernan, line 114: The rtol and maxsig here need further examination, should be coupled somehow. At 10 places, your relative error is at worst `5e-10`, which is under the default `rtol = 1e-9`. We stop short of `maxsig = 12`, yet the docstring says:

> Full-precision data (e.g. computed equivalent branches) reads `maxsig` figures

src/auto_tolerance.jl
luke-kiernan: I'm also a little cautious of the assumptions here: `1.0` becomes `1.0 - eps()` which then gets pretty printed as `0.999999`, and suddenly we've gained 6 figures of precision...

luke-kiernan (commented):
Interesting methodology. Things brought up by AI that seem worth mentioning:

1. Truncation shrinks towards 0, so errors here aren't random. This means that when summing a large number of entries, your error could compound instead of canceling.
2. Error amplification by downstream math. Woodbury correction in particular.
3. Since each row has its own sparsification threshold, global network invariants (like KCL) get funky. Similarly, if buses j and k are both distant from bus i, `PTDF[i,j]` and `PTDF[i,k]` could be both similarly small, yet one gets sparsified and the other doesn't. That gives you 100% error on their difference.

m-bossart (approved):

PR #329, opened 2026-06-18
Match PSS(e) zero-impedance branch handling (per-branch, r == 0)

### LINKED ISSUE #322: Zero impedance reduction handling
Comparing against PSS(e) on large networks, I'm noticing discrepancies due to ZIL handling. 

Currently, our [strategy](https://github.com/Sienna-Platform/PowerNetworkMatrices.jl/blob/main/src/apply_zero_impedance_reduction.jl#L47) is to calculate the ybus matrix, then go through off-diagonal elements `Y[i,j]`. If `real(Y[i,j]) == 0.0` and `abs(imag(Y[i,j]) >= susceptance_threshold`, then we merge the two buses. This diverges from PSS(e) in a few ways:
1. PSS(e) uses the L2 norm: `abs(y) >= threshold`. When I switch PNM to checking `abs(Y[i,j]) >= threshold`, our differences with PSS(e) drop significantly. Also, this just seems more numerically robust.
2. PSS(e) checks each branch, _not_ the combined Ybus entry. PSS(e) seems correct in this instance: _any_ zero impedance branch causes two buses to become electrically identified. Suppose we have two branches `b1`, `b2` in parallel with `abs(y_b1) >= threshold`, but `abs(y_b1 + y_b2) < threshold`. (Concretely: `y_b1 = 10^4 im` and `y_b2 = -10 im` would work.) An "examine `Y[i,j]` strategy" doesn't merge the buses; PSS(e)'s "examine each branch" does.
3. Similar to 2. above, but the opposite behavior: suppose we have two branches `b1`, `b2` in parallel with `abs(y_b1) < threshold` and `abs(y_b2) < threshold`, but `abs(y_b1 + y_b2) >= threshold`. In that scenario, an "examine `Y[i,j]` strategy" merges the buses; PSS(e)'s "examine each branch" doesn't.

Note: please correct me if I'm wrong about PSS(e)'s behavior. I'm pretty sure about (1) and (2) above, but I'd appreciate it if someone could verify PSS(e)'s behavior under scenario (3).

luke-kiernan:
Fixes #322.

## Problem

`ZeroImpedanceBranchReduction`'s `get_reduction` scanned the assembled `Ybus`
off-diagonals and merged a bus pair only when
`real(Y[i,j]) == 0 && abs(imag(Y[i,j])) >= susceptance_threshold`, testing the
*combined* parallel entry. So a zero-impedance jumper in parallel with a normal
line (whose combined `real(Y) != 0`) was not merged, even though PSS(e) merges it.

## Change

`get_reduction` now iterates the direct and parallel branch maps and tests each
branch individually: an arc is zero-impedance when **some individual
non-transformer branch has `r == 0` and `|y| >= susceptance_threshold`** —
matching PSS(e), which flags each line by its own impedance. A parallel group
whose members are each below threshold is not merged (no combined-entry term).

This keeps the per-branch examination (so the jumper-parallel-to-line case now
merges) while preserving PSS(e)'s `r == 0` gate: empirically, on a large PSS(e)
Ybus export every merged branch has `r == 0` and the largest unmerged `r == 0`
branch sits just under the threshold.

The merge `arc_key` is read directly as the branch-map key (identical to the old
reverse-lookup), so `removed_arcs` and the merge maps are unchanged. The loop
iterates only the direct and parallel maps; `series_branch_map` is empty when
ZIBR runs (it is always the first reduction).

## Tests

- Regression tests for the `r == 0` gate, the jumper-parallel case, and
  sub-threshold parallel groups (merged via neither measure).
- MATPOWER (`makeYbus`, unreduced) and the Eastern PSS(e) export (no
  zero-impedance section) are compared against the unreduced PNM matrix
  (`susceptance_threshold = Inf`); the other PSS(e) cases keep their
  zero-impedance groups and still validate that PNM matches PSS(e)'s merges.

### DIFF
--- a/src/Ybus.jl
+++ b/src/Ybus.jl
@@ -1511,7 +1511,22 @@ function _merge_ybus_buses!(
     end
     # Anti-parallel branches cancel in the signed adjacency (+1/-1 sum to 0) and get dropped
     # by `dropzeros!`, hiding a real edge from DegreeTwoReduction. The complex Ybus data sums
-    # and never cancels, so re-derive each surviving bus's adjacency from it.
+    # and never cancels structurally, so re-derive each surviving bus's adjacency from it.
+    _repair_merged_adjacencies!(adjacency_data, data, bus_lookup, merged_bus_pairs)
+    return
+end
+
+# Re-derive every surviving bus's signed adjacency from the complex Ybus data. Shared by the
+# in-place merge (`_merge_ybus_buses!`) and the fused pure-merge rebuild so the two paths
+# cannot drift. A merged off-diagonal can sum to a stored zero (e.g. a series capacitor
+# cancelling a line of equal magnitude), but that still marks a real edge between the buses,
+# so it must produce an adjacency entry; do not drop those zeros.
+function _repair_merged_adjacencies!(
+    adjacency_data::SparseArrays.SparseMatrixCSC{Int8, Int},
+    data::SparseArrays.SparseMatrixCSC{YBUS_ELTYPE, Int},
+    bus_lookup::Dict{Int, Int},
+    merged_bus_pairs::Dict{Int, Int},
+)
     for surviving_bus in Set(values(merged_bus_pairs))
         _repair_merged_adjacency!(adjacency_data, data, bus_lookup[surviving_bus])
     end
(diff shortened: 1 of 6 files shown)


luke-kiernan:
2 comments:

1.  Test failures. I'm guessing they're tests I need to adjust for changed behavior, not actual bugs, but will look into it.
2. Performance: ~50% longer ybus build time on both the EI and the 2k test system. Current implementation is doing structural insertions on the assembled sparse matrix (slow)--fixable, but requires more restructuring. Opening a stacked PR for it.

edit: first round of behavioral changes fixed

jd-lara (approved):

Issue #231, opened 2025-12-30
Handle PhaseShiftingTransformers in Parallel


m-bossart:
We currently skip parallel reduction of phase shifting transformers because depending on the parameters, we cannot necessarily derive an exact equivalent representation in the Ybus: 
https://github.com/NREL-Sienna/PowerNetworkMatrices.jl/blob/bc5a33ce6de08b3ee3e84934630287e855348ee2/src/definitions.jl#L13

However, with the current design, if we have parallel phase shifting transformers, we cannot have two entries in the direct_branch_map with the same key, therefore we over-write entries in `add_to_branch_maps!`:
https://github.com/NREL-Sienna/PowerNetworkMatrices.jl/blob/bc5a33ce6de08b3ee3e84934630287e855348ee2/src/Ybus.jl#L199

Issue #305, opened 2026-05-20
Bug: Branch silently dropped from `NetworkReductionData` when a `PhaseShiftingTransformer` is parallel to a regular branch


orennia-juan:
## Description

When a `PhaseShiftingTransformer` (PST) shares an arc with a `Line` or `Transformer2W`, one of the two branches is silently dropped from all `NetworkReductionData` maps in `add_to_branch_maps!` (`src/Ybus.jl`). The Ybus is built correctly (both admittances are summed), but the dropped branch is never passed to `set_power_flow!`, leaving its flow at zero after the solve.

## Root cause

`add_to_branch_maps!` routes each branch into `direct_branch_map` or `parallel_branch_map`. The condition that promotes two branches on the same arc to `parallel_branch_map` requires **both** to be outside `SKIP_PARALLEL_REDUCTION_TYPES`. When one is a PST (which is in `SKIP`), that condition is false, and the code falls through to the `else` clause, which unconditionally overwrites `direct_branch_map[arc_tuple]` with the incoming branch, losing the previous one.

| Step | Branch | Promoted to parallel? | Result |
|------|--------|-----------------------|--------|
| 1 | `Line` | — | `direct[arc] = Line` |
| 2 | `PST` | No — PST is in `SKIP` | `direct[arc] = PST` **(Line lost)** |

The same displacement occurs if the PST arrives first.

## Fix

Add an `elseif haskey(direct_branch_map, arc_tuple)` clause between condition B and the `else`. When the arc is already occupied and one branch is a SKIP type, move both into `parallel_branch_map` so neither is lost.

```julia
    # existing condition B (both branches outside SKIP)
    elseif haskey(direct_branch_map, arc_tuple) &&
           typeof(direct_branch_map[arc_tuple]) ∉ SKIP_PARALLEL_REDUCTION_TYPES &&
           typeof(br) ∉ SKIP_PARALLEL_REDUCTION_TYPES
        corresponding_branch = direct_branch_map[arc_tuple]
        delete!(direct_branch_map, arc_tuple)
        delete!(reverse_direct_branch_map, corresponding_branch)
        parallel_branch_map[arc_tuple] = BranchesParallel([corresponding_branch, br])
        reverse_parallel_branch_map[corresponding_branch] = arc_tuple
        reverse_parallel_branch_map[br] = arc_tuple
    # NEW: one branch is a SKIP type — still store both so neither is lost
[...]

PR #339, opened 2026-07-27
Jd/transformer refactor part 1

### LINKED ISSUE #201: Handle parallel branch reductions for branches with phase shift

jd-lara:
This is a first pass to consolidate transformers in PNM. This PR also fixes #201, #231 and #305


src/BranchAdmittance.jl
```diff
@@ -4,26 +4,22 @@
...
-# to 0 for non-shifting windings).
-_branch_tap(::PSY.ACTransmission) = 1.0
-_branch_tap(b::PSY.TwoWindingTransformer) = PSY.get_tap(PSY.get_circuit(b))
-_branch_shift(::PSY.ACTransmission) = 0.0
-_branch_shift(b::PSY.TwoWindingTransformer) = PSY.get_α(PSY.get_circuit(b))
-
 # Split a two-winding transformer's complex magnetizing shunt onto the π-model from/to shunt
 # slots per its `PSY.TwoWindingTransformerShuntLocation`: PRIMARY places the full value on
 # the from side, SECONDARY on the to side, SPLIT the full value on both sides.
 function _magnetizing_shunt_split(y_shunt::Complex, location)
```
m-bossart: add type to location

src/BranchAdmittance.jl
```diff
@@ -33,11 +29,15 @@ end
 function _three_winding_shunt_split(y_shunt::Complex, location, winding_number::Int)
```
m-bossart: same as above

src/BranchAdmittance.jl
```diff
@@ -62,56 +62,131 @@ fixed-ratio transformer has `tap = 1.0`, so this is a no-op for it and matches t
...
+    get_series_susceptance(PSY.get_circuit(t), units)
 
 """
-    branch_admittance(b::PSY.ACTransmission) -> NamedTuple
+    get_series_susceptance(c::PSY.TransformerCircuit, units::IS.AbstractUnitSystem)
 
-π-model admittance `(g, b, g_fr, b_fr, g_to, b_to, tap, shift)` for a line or symmetric
-branch, where `g + im*b == 1 / (r + im*x)` is the series admittance and the line charging
-susceptance is split between the from/to shunts.
+Tap-divided series susceptance of a transformer circuit. Both arities route here: the circuit
```
m-bossart: Unclear what "both" is referring to here. The method for ThreeWindingTransformerCircuit does not route here.

src/BranchAdmittance.jl
```diff
@@ -62,56 +62,131 @@ fixed-ratio transformer has `tap = 1.0`, so this is a no-op for it and matches t
...
+    b = PSY.get_b(br, PSY.SU)
+    return (complex(g.from, b.from), complex(g.to, b.to))
+end
+
+_get_shunts(::PSY.DiscreteControlledACBranch) = (zero(ComplexF64), zero(ComplexF64))
+
+# Kept out of line so the `@warn` body does not blow `equivalent_branch`'s inlining budget:
+# without this, the whole π chain stops inlining into `ybus_branch_entries` and the unit-tap
+# test in `_pi_to_ybus` can no longer constant-fold for lines.
+@noinline function _warn_zero_impedance(b, min_x_eps)
```
m-bossart: add types?

src/BranchAdmittance.jl
```diff
@@ -169,52 +291,47 @@ end
...
+comparison observes.)
 """
-function winding_admittance(w::ThreeWindingTransformerCircuit)
-    ys = 1.0 / (get_equivalent_r(w) + get_equivalent_x(w) * im)
-    sh = _three_winding_shunt_split(
-        PSY.get_magnetizing_shunt(get_transformer(w), PSY.SU),
-        PSY.get_shunt_location(get_transformer(w)),
-        get_winding_number(w),
-    )
+function _pi_to_ybus(adm)
```
m-bossart: add type

src/BranchAdmittance.jl
m-bossart: Specify both here means TwoWindingTransformer and ThreeWindingTransformerCircuit

src/BranchAdmittance.jl
```diff
@@ -62,56 +62,131 @@ fixed-ratio transformer has `tap = 1.0`, so this is a no-op for it and matches t
...
-        g_fr = 0.0, b_fr = b_sh.from, g_to = 0.0, b_to = b_sh.to,
-        tap = 1.0, shift = 0.0,
-    )
-end
+get_series_susceptance(c::PSY.TransformerCircuit, units::IS.AbstractUnitSystem) =
+    (1 / PSY.get_x(c, units)) / PSY.get_tap(c)
 
 """
-    branch_admittance(b::PSY.TwoWindingTransformer) -> NamedTuple
+    equivalent_branch(b; min_x_eps) -> EquivalentBranch
```
m-bossart: Is this doc string out of place? Should be below with the method?

src/BranchAdmittance.jl
```diff
@@ -62,56 +62,131 @@ fixed-ratio transformer has `tap = 1.0`, so this is a no-op for it and matches t
...
+
+# Kept out of line so the `@warn` body does not blow `equivalent_branch`'s inlining budget:
+# without this, the whole π chain stops inlining into `ybus_branch_entries` and the unit-tap
+# test in `_pi_to_ybus` can no longer constant-fold for lines.
+@noinline function _warn_zero_impedance(b, min_x_eps)
+    @warn "Branch $(PSY.get_name(b)) has r=0.0 and x=0.0; substituting x=$(min_x_eps) to avoid division by zero. This branch will be reduced by ZeroImpedanceBranchReduction unless its endpoints are irreducible."
+    return
+end
+
+@inline function equivalent_branch(
```
m-bossart, line 153: I understand why `_warn_zero_impedance` gets `@noinline`, but is there a reason for this method to use `@inline ` while the other `equivalent_branch` methods do not?

src/BranchAdmittance.jl
```diff
@@ -62,56 +62,131 @@ fixed-ratio transformer has `tap = 1.0`, so this is a no-op for it and matches t
...
+        PSY.get_magnetizing_shunt(transformer, PSY.SU),
+        PSY.get_shunt_location(transformer),
+        get_winding_number(w),
+    )
+    return _circuit_equivalent_branch(w.circuit, sh)
+end
+
+# The whole of "a transformer is a circuit": the circuit owns r/x/tap/α for both arities, so
+# 2W and 3W differ only in which shunt-split rule produced `sh`.
+function _circuit_equivalent_branch(circuit::PSY.TransformerCircuit, sh)
```
m-bossart: add type for `sh`

src/BranchAdmittance.jl
```diff
@@ -62,56 +62,131 @@ fixed-ratio transformer has `tap = 1.0`, so this is a no-op for it and matches t
...
     sh = _magnetizing_shunt_split(
         PSY.get_magnetizing_shunt(b, PSY.SU),
         PSY.get_shunt_location(b),
     )
-    return (
-        g = real(ys), b = imag(ys),
-        g_fr = sh.g_fr, b_fr = sh.b_fr, g_to = sh.g_to, b_to = sh.b_to,
-        tap = _branch_tap(b), shift = _branch_shift(b),
+    return _circuit_equivalent_branch(PSY.get_circuit(b), sh)
+end
```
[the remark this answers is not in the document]
m-bossart: I think this is correct and we should always apply `mix_x_eps` for all branches

src/BranchAdmittance.jl
```diff
@@ -62,56 +62,131 @@ fixed-ratio transformer has `tap = 1.0`, so this is a no-op for it and matches t
...
-π-model admittance for a reduction-aggregated arc (a `BranchesSeries` chain or
-`BranchesParallel` group), built from PNM's reduction-aware equivalent physical branch
-parameters. Series/parallel equivalents of lines carry `tap == 1`.
+π-model admittance `(g, b, g_fr, b_fr, g_to, b_to, tap, shift)` for any single branch, where
+`g + im*b == 1 / (r + im*x)` is the series admittance. The admittance-form view of
+[`equivalent_branch`](@ref); see it for the shunt and unit conventions.
 """
-function branch_admittance(segment, nr::NetworkReductionData)
-    eb = get_equivalent_physical_branch_parameters(segment, nr)
+function branch_admittance(b; min_x_eps::Float64 = ZERO_IMPEDANCE_X_EPSILON)
```
m-bossart: add type

src/BranchAdmittance.jl
```diff
@@ -121,22 +196,15 @@ function branch_admittance(segment, nr::NetworkReductionData)
...
-        shift = -adm.shift,
-    )
+"""
+    branch_admittance(segment, nr::NetworkReductionData) -> NamedTuple
+
+π-model admittance for a reduction-aggregated arc (a `BranchesSeries` chain or
+`BranchesParallel` group), built from PNM's reduction-aware equivalent physical branch
+parameters. Series/parallel equivalents of lines carry `tap == 1`.
+"""
+function branch_admittance(segment, nr::NetworkReductionData)
```
m-bossart: Should we add an abstract type that captures both the parallel and series branches.

test/test_arc_types_and_reductions.jl
```diff
@@ -128,7 +128,6 @@ end
             rating = 80.0,
             base_power = 100.0,
             base_voltage_primary = 230.0,
-            winding_group_number = WindingGroupNumber.GROUP_11,
```
m-bossart, line 131: This needs to be reverted

test/test_equivalent_getters.jl
```diff
@@ -58,6 +58,17 @@
     emergency_rating_eq = PNM.get_equivalent_emergency_rating(bp)
     @test emergency_rating_eq ≈ 250.0 atol = 1e-6
 
+    # A parallel group's normal-operation equivalent rating is the sum of its members: every
+    # circuit on the arc carries flow at once. The N-1 value is reserved for a parallel block
+    # embedded in a series chain (`_series_member_rating`).
+    @test PNM.get_equivalent_rating(bp) ≈ 250.0 atol = 1e-6
```
m-bossart, line 66: I thought the default was the `get_single_element_contingency_rating` version

test/test_equivalent_getters.jl
```diff
@@ -365,7 +376,6 @@ end
             base_power = 100.0,
             base_voltage_primary = 1.0,
             base_voltage_secondary = 1.0,
-            winding_group_number = WindingGroupNumber.GROUP_11,
```
m-bossart, line 368: revert

m-bossart:
This should also close #231 (there is some overlap between these issues)

.claude/CLAUDE.md
```diff
@@ -59,7 +59,11 @@ Exports live only in the main module file.
...
   - **Anti-parallel branches cancel in value-based adjacency** (+1/−1 sums to 0) — `_resolve_antiparallel_adjacency!` / `_repair_merged_adjacency!` restore connectivity; keep them in any new adjacency path.
+  - **3W circuits register through the merge-aware branch-map path.** `add_to_branch_maps!(::NetworkReductionData, ::PSY.ThreeWindingTransformer)` used to write each winding straight into `direct_branch_map[arc]`, silently overwriting anything already on that star-point arc (a `Line`, another winding, an existing group) and leaving a dangling `reverse_direct_branch_map` entry. It now delegates per circuit to the 3-arg `add_to_branch_maps!`, which promotes to a parallel group like any other branch. This was the 3W survival of the exact defect issue #231 described for 2W.
```
m-bossart: Is this needed? Can we not assume each 3WT has a unique star bus and that we won't end up with parallel paths that consist of 3WT circuits?

test/test_equivalent_getters.jl
m-bossart: After discussion, we decided to actually remove the WindingGroupNumber, so this comment no longer applies

src/Ybus.jl
```diff
@@ -279,12 +285,13 @@ function add_to_branch_maps!(
     parallel_branch_map = get_parallel_branch_map(nr)
     reverse_parallel_branch_map = get_reverse_parallel_branch_map(nr)
     arc_tuple = get_arc_tuple(arc, nr)
-    if haskey(parallel_branch_map, arc_tuple) && !_skip_parallel_reduction(br)
+    # Phase-shifting members are legitimate group members (the ZIR merge path already
```
m-bossart: I think these comments are duplicative with the implementation details in the docstring. Would suggest keeping the docstring and deleting the comment here

src/Ybus.jl
```diff
@@ -300,48 +307,38 @@ function add_to_branch_maps!(
...
 
-Three-winding transformers are modeled using a star (wye) configuration with three arcs
-connecting to a virtual star bus. Each available winding is mapped separately.
+Each circuit is filed through the same merge-aware path as any other `PSY.ACTransmission`
+(the 3-arg `add_to_branch_maps!`), so a winding whose star-point arc coincides with an
+already-registered branch (a `Line`, another winding, or an existing parallel group) is
+merged into a parallel group rather than silently overwriting the earlier entry.
+
+A unique star bus does not make that collision unreachable: arc keys resolve through
+`get_arc_tuple`, which remaps both endpoints, so two windings land on the same key as soon as
```
m-bossart: Good catch. I think the docstring is confusing / too strongly worded. "...so two winding circuits could land on the same key if a reduction merges the transformer's terminal buses."

src/PowerNetworkMatrix.jl
m-bossart, line 364: This function should have a docstring describing what it is used for (to scale matrix entries on retrieval). The name on its own is generic

src/common.jl
```diff
@@ -132,27 +92,27 @@ end
...
-function get_switched_admittances(sys::PSY.System, reverse_bus_search_map)
-    collection = Vector{PSY.SwitchedAdmittance}()
-    for sa in
-        collect(PSY.get_components(x -> PSY.get_available(x), PSY.SwitchedAdmittance, sys))
+# Available shunt components whose bus survived the reduction (a shunt on an eliminated bus
+# is already folded into its parent bus's diagonal).
+function _get_retained_shunts(
+    ::Type{T},
+    sys::PSY.System,
+    reverse_bus_search_map,
```
m-bossart: add type

src/common.jl
```diff
@@ -132,27 +92,27 @@ end
...
-    collection = Vector{PSY.FixedAdmittance}()
-    for sa in
-        collect(PSY.get_components(x -> PSY.get_available(x), PSY.FixedAdmittance, sys))
-        if !haskey(reverse_bus_search_map, PSY.get_number(PSY.get_bus(sa)))
-            push!(collection, sa)
-        end
-    end
-    return collection
-end
+get_switched_admittances(sys::PSY.System, reverse_bus_search_map) =
```
m-bossart: add type

src/common.jl
```diff
@@ -132,27 +92,27 @@ end
...
+    _get_retained_shunts(PSY.SwitchedAdmittance, sys, reverse_bus_search_map)
+
+get_fixed_admittances(sys::PSY.System, reverse_bus_search_map) =
```
m-bossart: add type

src/NetworkReductionData.jl
```diff
@@ -491,6 +491,9 @@ Gets the concrete types of all AC transmission branches included in an instance
 _ac_transmission_type(x::PSY.ACTransmission) = typeof(x)
 _ac_transmission_type(w::ThreeWindingTransformerCircuit) = get_transformer_type(w)
 
+_is_transformer_circuit(::PSY.ACTransmission) = false
```
m-bossart: This name is confusing because TransformerCircuit now refers to the data structure that is common to both TwoWindingTransformer and ThreeWindingTransformer, but this is specifically for the ThreeWindingTransformerCircuit in PNM

src/NetworkReductionData.jl
```diff
@@ -576,7 +579,7 @@ function Base.show(io::IO, ::MIME{Symbol("text/plain")}, nrd::NetworkReductionDa
         "\tNumber of series arcs (number of branches): $(length(nrd.series_branch_map)) ($(length(nrd.reverse_series_branch_map)))",
     )
     println(
```
m-bossart: This show was from when the 3WT arcs were in a separate map. I think it probably makes sense to just delete it and only report the direct/series/parallel (any of which count contain 3wt arcs)

m-bossart (approved):

PR #340, opened 2026-07-30
multi ref bus

jd-lara:
These are other fixes for the multiref

### DIFF
--- a/src/AdjacencyMatrix.jl
+++ b/src/AdjacencyMatrix.jl
@@ -52,8 +52,6 @@ end
 get_axes(M::AdjacencyMatrix) = M.axes
 get_lookup(M::AdjacencyMatrix) = M.lookup
 get_ref_bus(M::AdjacencyMatrix) = sort!(collect(keys(M.subnetwork_axes)))
-get_ref_bus_position(M::AdjacencyMatrix) =
-    [get_bus_lookup(M)[x] for x in keys(M.subnetwork_axes)]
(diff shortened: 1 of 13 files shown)


m-bossart (commented):
Does it still make sense to have a function `get_ref_bus` that doesn't consider multiple ref buses in the same sychronous subsystem?

jd-lara:
> Does it still make sense to have a function `get_ref_bus` that doesn't consider multiple ref buses in the same sychronous subsystem?

Good question, my inclination is that we should make the behavior more concrete and if there are multiple Swing buses in a case, only one is a ref and the rest are slacks and the model for slacks -> fixed V and \theta and all refs behave like slacks but we need to validate that with PSSe

m-bossart:
> > Does it still make sense to have a function `get_ref_bus` that doesn't consider multiple ref buses in the same sychronous subsystem?
> 
> Good question, my inclination is that we should make the behavior more concrete and if there are multiple Swing buses in a case, only one is a ref and the rest are slacks and the model for slacks -> fixed V and \theta and all refs behave like slacks but we need to validate that with PSSe

Doesn't this have to be true given PSSE only has PV,PQ, Swing,Disconnected bus types. They they can't distinguish between refs/slacks with just these designations.

m-bossart (approved):
I'm ok with these changes if we want to save the larger discussion about the semantics of ref/slack/swing for another change

PR #341, opened 2026-07-31
Partition parallel branches deterministically

acostarelli:
It seems that binning branches by their impedance angle is affected by the order in which you read branches / create bins. This approach updates the "representative" bin angle with the bin median. This is probably not the most efficient approach, but I don't know if this is a hot-spot.

I also don't know if this would've ever caused an error. I would imagine if anything this results in fewer bins, but maybe also solves issues that might stem from reading the same vector of branches but in a different order (not sure if that would ever happen though).


acostarelli:
Closed because I messed up, but it didn't throw me an error so I'll look into why.

PR #350, opened 2026-08-27
Mb/network flow

m-bossart:
One source of truth for the reduced network's branch index.

  The per-type branch index used to be filled lazily by populate_branch_maps_by_type!,
  and only POM ever called it — so the index was invisible to PowerFlows, and POM had to
  deepcopy the reduction to avoid state leaking between builds. BranchCatalog now wraps
  NetworkReductionData and is built when the matrix is built, so the index exists
  because the matrix exists. Optimization-side branch filters become a predicate applied
  when deriving a catalog, instead of mutating shared reduction state.

  Also fixes two defects found on the way: a 3W winding promoted into a parallel group
  was indexed under a key its accessor never looked at, and WardReduction silently
  widened its study area until it covered the whole network, making the reduction a
  no-op that still reported success.


src/BranchCatalog.jl
```diff
@@ -0,0 +1,268 @@
+const ARC_ENTRY = Tuple{Tuple{Int, Int}, Symbol}
```
luke-kiernan, line 1: I try to avoid using symbols because they provide both the human reader and the compiler nothing to go off of. Looks like you're using it to tell which map holds the arc. Switch to `Ref{Any}` or `Ref{AbstractDict}`?

src/BranchCatalog.jl
```diff
@@ -0,0 +1,268 @@
...
+const NAME_TO_ARC = Dict{DataType, DataStructures.SortedDict{String, ARC_ENTRY}}
+const COMPONENT_TO_ENTRY = Dict{DataType, Dict{String, String}}
+const COMPONENT_NAME_INDEX = Dict{String, Vector{Tuple{DataType, Tuple{Int, Int}, Symbol}}}
```
luke-kiernan, line 4: Can you use `Dict` instead of a `Vector` here as the value type? Names are unique per-type, so it'd seem like you should be able to.

src/BranchCatalog.jl
```diff
@@ -0,0 +1,268 @@
...
+    BranchCatalog
+
+An immutable, per-branch-type index over the branch maps a network reduction produced. Every
+matrix carries one, reachable with `get_branch_catalog`.
+
+# Fields
+- `network_reduction_data::NetworkReductionData`: the reduction this indexes
+- `maps_by_type::BranchMapsByType`: the six reduction maps re-bucketed by branch type
+- `name_to_arc`: per type, entry name => (arc, which map holds it)
+- `component_to_entry_name`: per type, component name => name of the entry representing it
```
luke-kiernan, line 20: What are "entries?" Post-reduction elements, not necessarily present in the final network (e.g. because nested reductions, like parallel degree 2 reductions)?

Similar for candidate below: candidate what?

src/BranchCatalog.jl
```diff
@@ -0,0 +1,268 @@
...
+##############################################################################
+############################ Entry matching ##################################
+##############################################################################
+
+_keep_all(::Type, ::Any) = true
+
+_is_unfiltered(predicate) = predicate === _keep_all
+
+function _warn_mixed_group(kind::String, branches)
+    @warn "$kind contains mixed branch types, filters might be applied to more " *
```
luke-kiernan, line 75: Might be worth adding a `maxlog` on the warning here

luke-kiernan (commented):
Can you iterate with AI to compose some human readable summary of what's going on here? I'm getting lost in the layers of bookkeeping and composition.

Before, we basically had arc in reduced network corresponds one-to-one with a reduced network element, which in turn was an entry in some dictionary, depending on the type of reduction that created the element.

Aside: one idea would be to switch from storing "how to find the thing" to "ref to the thing." Might be rough type-stability wise, though

docs/Project.toml
```diff
@@ -5,10 +5,17 @@ Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
 DocumenterInterLinks = "d12716ef-a0f6-4df4-a9f1-a5a34e75c656"
 InfrastructureSystems = "2cd47ed4-ca9b-11e9-27f2-ab636a7671f1"
 Literate = "98b081ad-f1c9-55d3-8b20-4c87d4299306"
+PowerCoreOpenAPIModels = "b7b40286-e793-417d-a9a0-b1583e4da1cb"
+PowerDynamicsOpenAPIModels = "044a0b22-31f8-4ef6-8282-c9a61b3013f6"
```
jd-lara, line 9: We don't need the dynamics PowerOpenAPI in this one

jd-lara (changes requested):
Directionaly this looks ok to me, Luke's comments make sense

PR #353, opened 2026-08-31
feat: expose a chain's segment orientations, replacing _get_chain_data

jd-lara:
`_get_chain_data(equivalent_arc, chain, nrd)` was deleted in `e67ee3b`, but PowerOperationsModels still calls it to sign interface flow contributions for a degree-two chain — the one PNM symbol of the 63 it references that no longer resolves, and an `UndefVarError` waiting on any `TransmissionInterface` whose direction mapping covers a folded chain.

The information it computed is now stored: `_build_chain_segments!` walks a chain from `arc_key[1]` to `arc_key[2]` and `add_branch!` records each segment's orientation as it goes, which is the same walk the deleted function performed. So the stored vector equals what it returned, and no recomputation is needed.

`get_segment_orientations(bs)` returns it. The two-argument method takes the caller's equivalent arc and errors unless it is the chain's own `arc_key`. That check is the deleted function's contract, not a new restriction: it walked segments in stored order from `equivalent_arc[1]`, so a reversed arc threw "Found disconnected series chain" rather than returning flipped orientations. Erroring keeps that rather than guessing a reversal — which matters because a chain inside a `BranchesParallel` keeps its own key while the group is framed on the principal one, so the two need not agree.

`test_chain_orientation_frame.jl` pins the equivalence against the deleted algorithm, kept verbatim as the test's reference. It covers standalone chains, a chain nested in a parallel group, and `build_reversed_asymmetric_degree_two_chains`, where the group is framed (1,3) and a member is keyed (3,1): that member agrees with the reference in its own frame and rejects the group's.

`get_arc_key` accompanies it, and the three sites that read the field directly now go through the accessor.

### DIFF
--- a/test/test_chain_orientation_frame.jl
+++ b/test/test_chain_orientation_frame.jl
@@ -1,11 +1,8 @@
 #=
-`PNM._get_chain_data(equivalent_arc, chain, nrd)` was deleted in `e67ee3b`; PowerOperationsModels
-still calls it to sign interface flow contributions. `get_segment_orientations` replaces its
-second return value. These tests pin the equivalence, because the replacement reads a stored
-field where the old code recomputed the walk.
-
-The reference below is the removed function's body verbatim. It lives here, not in `src/`: the
-point is to prove the stored vector agrees with it, not to keep the algorithm.
+`get_segment_orientations` replaces the second return value of the removed `_get_chain_data`,
+which PowerOperationsModels still calls to sign interface flow contributions. The reference
+below is that function's body verbatim — the point is to prove the stored vector agrees with
+it, not to keep the algorithm.
 =#
 
 function _reference_chain_data(equivalent_arc, chain, nrd)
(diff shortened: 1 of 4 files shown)


test/test_chain_orientation_frame.jl
```diff
@@ -0,0 +1,149 @@
...
+        # The map key is the chain's own frame, which is what the old caller passed.
+        @test PNM.get_arc_key(chain) == arc
+        @test PNM.get_segment_orientations(chain) ==
+              _reference_chain_data(arc, chain, nrd)[2]
+        @test PNM.get_segment_orientations(chain, arc) ==
+              _reference_chain_data(arc, chain, nrd)[2]
+    end
+end
+
+@testset "get_segment_orientations matches for chains nested in a parallel group" begin
```
m-bossart: The title of this test makes it seem like it is testing that the segments match for both sides of the parallel group, which is not necessarily true. In reality it is doing the same as above (testing get_setment_orientations matches the removed _get_chain_data walk) when the chains are also independently part of a parallel group. Suggest changing title to clarify

test/test_chain_orientation_frame.jl
```diff
@@ -0,0 +1,149 @@
+#=
```
m-bossart, line 1: Verbose comment, shouldn't need to reference the specific commit when the function was removed.

m-bossart (approved):
Two very minor comments, otherwise looks good

PR #356, opened 2026-08-31
Name reduced arcs for the arc, not for their members

luke-kiernan:
Names a reduced arc for the arc it occupies instead of for its members.

## The problem

A composite's name was the longest common prefix of its member names plus a suffix. That is not injective — `La ∥ Lb` and `Lc ∥ Ld` both produce `Ldouble_circuit` — so two groups could collide under one name. It also moved whenever membership did: adding a circuit to a corridor renamed it, and results stopped comparing across reduction settings.

The degenerate cases are worse than the collision. A prefix can be empty, or can be a bare bus number: POM's branch-rating fixture was reaching a group named `1double_circuit`.

## The change

Composites are named for their arc:

```julia
get_name(bp::AbstractBranchesParallel) = "$(bp.arc_key[1])_$(bp.arc_key[2])_double_circuit"
get_name(bs::BranchesSeries)           = "series_$(bs.arc_key[1])_$(bs.arc_key[2])"
```

`_composite_entries` already admits at most one composite per unordered bus pair, so `arc => name` is injective by construction.

**The `double_circuit` suffix is kept deliberately.** It is what tells a reader of results that the row is a total across the parallel members rather than one member's own flow. Only the stem changed. `_longest_starting_substring` is deleted.

Direct arcs keep their component's name, so the break lands only on names that were unstable anyway.

## Scope

Naming only — no change to which rows exist. A series chain still reports one row per segment under the segment's own component name, because each segment carries the arc's flow; a parallel group still reports once. `series_<from>_<to>` is the chain's identity in the catalog, never a reporting row, which is why the suffix question doesn't arise there.

## Breaking

Composite result keys change: `1-4-i_double_circuit` → `1_4_double_circuit`. Any fix to the old scheme breaks compatibility, since the old names were the defect.

## Testing

Suite 10,889,477. Formatter clean.

Also merges `psy6` and adds the two packages its OpenAPI-split commit put in `[deps]` to Aqua's stale-deps ignore list — pin-only entries, exactly as the comment above that list already describes. Without it `Aqua.test_stale_deps` fails on `psy6` itself and on anything branched from it; if `psy6` fixes this directly, expect a trivial conflict on those three lines.

### DIFF
--- a/src/BranchCatalog.jl
+++ b/src/BranchCatalog.jl
@@ -86,6 +86,28 @@ end
 _entry_matches(device::T, predicate) where {T <: PSY.ACTransmission} =
     predicate(T, device)
 
+"""
+    _entry_name(arc, entry) -> String
+
+The name an arc's entry is indexed under.
+
+A direct arc is one branch and keeps that branch's own name: it is already unique per type,
+already stable, and renaming it would move result keys for the ~90% of arcs no reduction
+touched. A composite arc takes its name from the arc instead.
+
+Deriving a composite's name here rather than on the aggregate is deliberate on two counts.
+The catalog knows the arc the entry is *filed under*, while an aggregate's `arc_key` holds
+original bus numbers that `reverse_bus_search_map` may since have remapped -- the two can
+disagree. And `_composite_entries` admits at most one composite per unordered bus pair, so
+arc => name is injective by construction here, where on the aggregate it was the longest
+common prefix of member names: `La`/`Lb` and `Lc`/`Ld` both yielded `Lseries_chain`, and the
+name moved whenever membership did.
+"""
+_entry_name(::Tuple{Int, Int}, entry::PSY.ACTransmission) = get_name(entry)
+_entry_name(arc::Tuple{Int, Int}, ::AbstractBranchesParallel) =
+    "parallel_$(arc[1])_$(arc[2])"
+_entry_name(arc::Tuple{Int, Int}, ::BranchesSeries) = "series_$(arc[1])_$(arc[2])"
+
 ##############################################################################
 ############################## Index building ################################
 ##############################################################################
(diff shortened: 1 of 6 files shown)


src/BranchCatalog.jl
```diff
@@ -86,6 +86,28 @@ end
...
+The catalog knows the arc the entry is *filed under*, while an aggregate's `arc_key` holds
+original bus numbers that `reverse_bus_search_map` may since have remapped -- the two can
+disagree. And `_composite_entries` admits at most one composite per unordered bus pair, so
+arc => name is injective by construction here, where on the aggregate it was the longest
+common prefix of member names: `La`/`Lb` and `Lc`/`Ld` both yielded `Lseries_chain`, and the
+name moved whenever membership did.
+"""
+_entry_name(::Tuple{Int, Int}, entry::PSY.ACTransmission) = get_name(entry)
+_entry_name(arc::Tuple{Int, Int}, ::AbstractBranchesParallel) =
+    "parallel_$(arc[1])_$(arc[2])"
```
jd-lara, line 108: This could be a bottle neck on performance. We use this https://github.com/joshday/StringTemplates.jl/ in IS worth opening an issue about

src/BranchesParallel.jl
```diff
@@ -76,35 +76,24 @@ function _is_phase_shifting(bp::AbstractBranchesParallel)
...
+group was formed; a later reduction can remap them, so this is an identity for reading, not a
+key to look the group up with. `BranchCatalog._entry_name` derives the *indexed* name from
+the arc the catalog files the entry under, which is the remapped one.
+
+The prefix stem was not injective -- `La` ∥ `Lb` and `Lc` ∥ `Ld` both produced
+`Ldouble_circuit` -- and it moved whenever membership did, so a corridor was renamed by the
+act of adding a circuit to it. The arc stem is injective by construction.
+"""
+get_name(bp::AbstractBranchesParallel) =
+    "$(bp.arc_key[1])_$(bp.arc_key[2])_double_circuit"
```
jd-lara, line 96: Same here

jd-lara (approved):
The PR overall looks ok. It is hard to know how many parallel circuits we could find. I'll merge and @luke-kiernan can you open an issue to use https://github.com/joshday/StringTemplates.jl/ in those names please.

PR #357, opened 2026-08-31
Give a reduced arc one identity, keyed by the arc

luke-kiernan:
Builds on #356. Gives a reduced arc one identity — the arc — and deletes the map-of-origin tag.

## The problem

An arc had three identities and none was the arc:

- **name** — a member-name prefix (#356 fixes this)
- **location** — a `Symbol` naming which of six internal maps held the entry, handed to consumers so they could navigate *back* to an entry the catalog already had
- **structure** — rebuilt by separate forward and reverse walks that could disagree. Twice they did: PNM #351, and again on this branch when one walk was converted and the other wasn't.

## The change

```julia
struct ArcEntry
    entry::PSY.ACTransmission
    name::String
    leaves::Vector{PSY.ACTransmission}
end
```

**One arc-keyed table.** `arcs::Dict{arc, ArcEntry}` holds the entry, its name and its leaves; every other index reads them back rather than recomputing, so the projections can't disagree.

**The tag is deleted.** `get_reduction_entry(catalog, arc)` returns the entry directly. The stored value drops from `Tuple{Tuple{Int,Int}, Symbol}` to `Tuple{Int,Int}` — concrete *and* isbits.

**Provenance is a dispatch, not a field.** `arc_provenance(entry)` reads it off the entry's type: `DirectArc`, `ParallelArc`, `SeriesArc`, `SyntheticArc`. Aggregates subtype `PSY.ACTransmission`, so each gets its own arm rather than falling into a blanket method. `get_branch_multiplier` goes from twelve lines to three, each provenance stating its own reason instead of everything non-direct landing in the parallel path.

## What this does *not* change

The reporting shape. `name_to_arc`'s keys are the rows results are reported under, which is a different question from what an arc is called. A lossless chain carries the same flow in every segment, so each segment stays a row under its own component name; a parallel group reports once because its members' individual flows are never computed. That rule now applies at every depth — a parallel group inside a chain is one row.

An earlier revision of this branch flattened that asymmetry to one row per arc, on the mistaken reading that it was an inconsistency. It is the physics. Measured on ACTIVSg10k, collapsing chains drops 691 result rows to 332 — 359 physical lines losing their column.

**The whole suite passed both before and after that regression.** Nothing pinned the shape, which is how it got as far as it did. `test_branch_catalog.jl` now pins it across all three depths, plus the invariant that every redirect target is itself a row. Verified by mutation: filing segments under the chain name fails it at six assertions.

## Testing

Suite 10,889,526. Formatter 0, docs 0.

## Downstream

POM #260 adopts this. Merge #356 first.

### DIFF
--- a/src/BranchCatalog.jl
+++ b/src/BranchCatalog.jl
@@ -1,7 +1,31 @@
-const ARC_ENTRY = Tuple{Tuple{Int, Int}, Symbol}
+"""
+One reduced arc: the entry occupying it, the name it is indexed under, and every physical
+branch at its leaves.
+
+`leaves` is the only derived thing worth storing. It is what every type-keyed lookup needs
+and it costs a tree walk to recompute, whereas the *structure* is already reachable by
+iterating `entry`, and the *provenance* is already `entry`'s type -- see
+[`arc_provenance`](@ref). Empty leaves means an arc backed by no component, which today only
+a Ward equivalent can be.
+"""
+struct ArcEntry
+    entry::PSY.ACTransmission
+    name::String
+    leaves::Vector{PSY.ACTransmission}
+end
+
+get_entry(e::ArcEntry) = e.entry
+get_name(e::ArcEntry) = e.name
+get_leaves(e::ArcEntry) = e.leaves
+
+# The arc alone: concrete and isbits, 16 bytes. This carried a `Symbol` naming the
+# `BranchMapsByType` field that held the entry, so consumers could navigate back to it; that
+# is now `get_reduction_entry`, and the arc is the whole identity.
+const ARC_ENTRY = Tuple{Int, Int}
+const ARC_TABLE = Dict{ARC_ENTRY, ArcEntry}
 const NAME_TO_ARC = Dict{DataType, DataStructures.SortedDict{String, ARC_ENTRY}}
 const COMPONENT_TO_ENTRY = Dict{DataType, Dict{String, String}}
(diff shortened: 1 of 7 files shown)


src/BranchCatalog.jl
```diff
@@ -105,9 +160,96 @@ name moved whenever membership did.
...
+arc_provenance(c::BranchCatalog, arc::ARC_ENTRY) =
+    arc_provenance(get_reduction_entry(c, arc))
+
+"""
+    _branch_multiplier(provenance, entry, branch_name, arc) -> Float64
+
+Factor scaling a per-arc matrix entry to the named branch's share of it, dispatched on how
+the arc came to exist. Backs [`get_branch_multiplier`](@ref).
+"""
+_branch_multiplier(::DirectArc, ::PSY.ACTransmission, ::AbstractString, ::ARC_ENTRY) = 1.0
```
jd-lara, line 213: I don't follow what the purpose of this method is

jd-lara (commented):
I have one comment but generally I like the whole Provenance thing.

jd-lara (approved):

src/BranchCatalog.jl
luke-kiernan: We call `get_branch_multiplier` when we want to calculate per-component flows from an arc flow. That function used to be backed by if/else logic; now it's backed by multiple dispatching `_branch_multiplier`.

For a non-reduced arc (`::DirectArc`), 100% of the flow goes thru the one and only system component that corresponds to that arc.

PR #360, opened 2026-09-03
Let reverse indexing defer to the forward filter verdict

luke-kiernan:
Follow-up to #357, addressing Copilot's review comment on `_index_reverse!`. The finding is real; this is the guard plus a regression test.

## The bug

`_entry_matches` is not uniform across aggregates:

- `BranchesParallel` → `any(member matches)`
- `MixedBranchesParallel` → **`all(member matches)`**
- `BranchesSeries` → `all(segment matches)`

Suppose you have 2 components of different types in parallel (a `MixedBranchesParallel`), where one passes the filter and one doesn't.

Forward indexing goes thru the components asking about the **group**: under `all`, it skips the whole. As we do the forward index, we assemble `arcs`, so the arc for the parallel combined arc isn't in `arcs`.

Reverse indexing goes thru the components asking about the **members**: it skips the components that doesn't fit the filter, but then grabs the one that passes. It tries to get its name via `get_name(arcs[combined_arc])` -> `KeyError`.

Worth noting the old code was not correct here either. It did not throw, but it wrote a redirect into `component_to_entry` naming an entry that is not a row in `name_to_arc` — a dead end for any caller following it. The `KeyError` is that silent dangling pointer becoming loud.

## The fix

`arcs` is the forward pass's output, and the reverse pass consumes it as an input. A miss is a decision, not an accident:

```julia
haskey(arcs, arc) || continue
```

`BranchesSeries` shares the `all` asymmetry and is unaffected only because `_index_reverse_series!` takes no `arcs` and writes no redirect name — it would acquire the same defect the moment series arcs get the redirect treatment the other two have.

## Testing

New testset forces a `Line` and a `TwoWindingTransformer` onto one arc, which promotes them to a `MixedBranchesParallel`, then filters `Line` out. It asserts the catalog builds, the group is a row under neither type, neither member redirects to it — the transformer included, even though it matched — and the general invariant that every redirect names a row.

The unfiltered case is asserted first, so the filtered assertions cannot pass vacuously on an empty catalog. Verified by mutation: dropping the guard fails the testset with `KeyError: key (9, 14) not found`.

Suite 10,889,535. Formatter 0, docs 0.

## Base

Targets `lk/arc-derived-entry-names` rather than `psy6`, since that is where #357 landed and `psy6` does not carry `ArcEntry` yet.

https://claude.ai/code/session_017XUcz3ZFnWmRgwgrDjKsBb

### DIFF
--- a/src/BranchCatalog.jl
+++ b/src/BranchCatalog.jl
@@ -322,6 +322,12 @@ function _index_reverse!(
 )
     for (member, arc) in source
         _entry_matches(member, predicate) || continue
+        # The forward pass owns `arcs`, and its verdict is the catalog's verdict. Reaching a
+        # different one here is not possible for a group matching on `any` -- acceptance
+        # there is a superset of its members' -- but `MixedBranchesParallel` matches on
+        # `all`, so a member can pass this predicate while its group did not. Redirecting it
+        # would name an entry that is not a row.
+        haskey(arcs, arc) || continue
         T = _get_segment_type(member)
         _store!(get!(() -> Dict{typeof(member), Tuple{Int, Int}}(), dest, T), member, arc)
         # The name comes from the table, not from a second call to `_entry_name`: one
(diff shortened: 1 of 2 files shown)


jd-lara (approved):
Makes sense

PR #362, opened 2026-09-03
fix pin irreducible buses bug

m-bossart:


jd-lara (approved):
Seems straightforward, I'll merge once the CI/CD is green

PR #364, opened 2026-09-09
Made DC path respect minimum reactance

m-bossart:
## Problem
  
  A branch with `r == x == 0` has series susceptance `1/x == Inf`. Ybus and BA assembly substitute the
  reduction's `minimum_retained_impedance` for such a branch; the component-side accessors read the stored
  reactance and returned `Inf`. The two silently disagreed.

  Three measured consequences:

  - A full outage of a zero-impedance branch produced `delta_y12 = 0.0` — the contingency did nothing.
  - PTDF member weights came out `[10.0, Inf]`, normalizing to `NaN`.
  - `BA_Matrix` gave a zero-impedance phase shifter zero DC coupling while Ybus held a finite admittance for
  it — two answers in one matrix, twelve lines apart.

  Not a corner case: `PSY.TransformerCircuit` defaults `r = x = 0`, and ZIR excludes transformer arcs, so
  such a branch survives with no pinning.

  ## Change

  `get_series_susceptance` becomes three layers:

  - `_series_susceptance_raw` — the stored value, may be `Inf`, internal.
  - `get_effective_series_susceptance(segment, nr)` — substitutes the configured epsilon. Exported, and what
  every internal consumer now calls.
  - `get_series_susceptance` — the stored value; now **throws** naming the branch instead of returning `Inf`.

  `BranchesSeries` moves to impedance space (`1/Σ x·tap` rather than `1/Σ inv(b)`), removing the one place
  `Inf` was load-bearing and making the assertion possible.

  A property sweep pins the invariant across fixtures, so a future consumer that reads the raw value fails
  loudly on its first degenerate case instead of poisoning a matrix.

  ## Downstream

  POM needs one line (`RepresentativeBranch._dc_susceptance` → the new accessor). PF needs nothing — it has
  no references to any changed symbol.

  ## Testing

  - PNM: 10,889,617 pass, 0 fail.
  - POM full suite: 128,404 pass against this branch.
  - PF: 1457 pass / 3 fail — the 3 are pre-existing bit-exact float comparisons in
  `test_discrete_control.jl`, confirmed by re-running against a pre-branch worktree.


src/common.jl
```diff
@@ -414,6 +414,53 @@ function _oriented_member_phase_shift(
...
+_zero_impedance_susceptance(c::PSY.TransformerCircuit, min_x_eps::Float64) =
+    (1 / min_x_eps) / PSY.get_tap(c)
+_zero_impedance_susceptance(bp::AbstractBranchesParallel, min_x_eps::Float64) =
+    sum(_finite_series_susceptance(br, min_x_eps) for br in bp)
+_zero_impedance_susceptance(bs::BranchesSeries, min_x_eps::Float64) =
+    1 / sum(inv(_finite_series_susceptance(seg, min_x_eps)) for seg in bs)
+
+# Kept as a guard on the raw layer rather than a parallel implementation of it, so every
+# non-degenerate branch keeps the single source of truth.
+function _finite_series_susceptance(segment, min_x_eps::Float64)
```
m-bossart: add type here (and below)

src/common.jl
```diff
@@ -422,13 +469,23 @@ Susceptance-weighted equivalent DC phase shift of a parallel group,
...
+        α = _oriented_member_phase_shift(br, bp, nr)
+        b = _finite_series_susceptance(br, min_x_eps)
         b_total += b
-        b_alpha += b * _oriented_member_phase_shift(br, bp, nr)
+        if !iszero(α)
+            shifted = true
+            b_alpha += b * α
+        end
     end
+    # Short-circuit rather than divide when no member shifts: b_alpha is 0.0 in that case,
```
m-bossart: unneccesary comment; self explanatory from code

src/common.jl
```diff
@@ -414,6 +414,50 @@ function _oriented_member_phase_shift(
...
+# uses in Ybus assembly. Such branches normally merge away, but a pair between two
+# irreducible buses survives as a `BranchesParallel`.
+_zero_impedance_susceptance(::PSY.ACTransmission, min_x_eps::Float64) = 1 / min_x_eps
+_zero_impedance_susceptance(t::PSY.TwoWindingTransformer, min_x_eps::Float64) =
+    _zero_impedance_susceptance(PSY.get_circuit(t), min_x_eps)
+_zero_impedance_susceptance(w::ThreeWindingTransformerCircuit, min_x_eps::Float64) =
+    _zero_impedance_susceptance(w.circuit, min_x_eps)
+# Only the susceptance form is tap-divided, matching `get_series_susceptance`.
+_zero_impedance_susceptance(c::PSY.TransformerCircuit, min_x_eps::Float64) =
+    (1 / min_x_eps) / PSY.get_tap(c)
```
luke-kiernan, line 427: Just want to verify that this is intentional, not just going along with AI. (My gut instinct would be to omit the tap, because `min_x_exps` is really a mangitude, not a vector quantity.)

src/common.jl
```diff
@@ -414,6 +414,50 @@ function _oriented_member_phase_shift(
     return α
 end
 
+# The substituted susceptance for an `r == x == 0` branch, matching what `equivalent_branch`
+# uses in Ybus assembly. Such branches normally merge away, but a pair between two
```
luke-kiernan, line 418: > matching what `equivalent_branch` uses in Ybus assembly

would be convenient to have a common provider, then, instead of two different things to keep in sync.

src/common.jl
```diff
@@ -936,41 +997,44 @@ Returns 0.0 if the segment (or all branches in a parallel group) is fully trippe
 function _segment_susceptance_after_outage(
     segment::PSY.ACTransmission,
     tripped_set::Set{<:PSY.ACTransmission},
+    nr::NetworkReductionData,
```
luke-kiernan: Looks like you only need `nr::NetworkReductionData` for `_minimum_retained_impedance(nr)`: might be conceptually clearer to just pass that float, instead of `nr` as a whole.

src/definitions.jl
```diff
@@ -21,6 +21,10 @@ reduction-aware `(segment, nr)` form instead.
 """
 abstract type AbstractReductionAggregate <: PSY.ACTransmission end
 
+# `PSY.TransformerCircuit` is a `DeviceParameter`, not an `ACTransmission`, so anything
+# carrying a series impedance spans two unrelated supertypes.
+const SeriesSegment = Union{PSY.ACTransmission, PSY.TransformerCircuit}
```
luke-kiernan: > `PSY.TransformerCircuit` is a `DeviceParameter`, not an `ACTransmission`

That's rather awkward...`TransformerCircuit` a transmission element in practice, but our type hierarchy says it isn't one. We should keep an eye out for similar issues downstream, places where `ACTransmission` really ought to be this union.

src/ThreeWindingTransformerCircuit.jl
```diff
@@ -79,11 +79,23 @@ follows `x`: a star-leg reactance can legitimately be negative, giving a negativ
...
 [`get_equivalent_x`](@ref), which is pinned to the system base because reduction
 aggregation must sum on a common base.
+
+Throws if the star-leg reactance is zero (susceptance is non-finite). A consumer that needs
+the value the matrices actually use should call `get_effective_series_susceptance` instead.
 """
-get_series_susceptance(
+function get_series_susceptance(
+    segment::ThreeWindingTransformerCircuit,
+    units::IS.AbstractUnitSystem,
```
luke-kiernan: > `units::IS.AbstractUnitSystem`

I'd replace with type parameter for perf reasons. (It might compile the same either way, though.)

src/BranchesSeries.jl
```diff
@@ -187,10 +187,31 @@ function get_series_susceptance(
...
-        sum(inv(get_series_susceptance(x, units)) for x in series_chain)
-    total_susceptance = 1 / series_susceptances_sum
-    return total_susceptance
+    v = _series_susceptance_raw(series_chain, units)
+    isfinite(v) || _throw_non_finite_susceptance(series_chain, v)
+    return v
+end
+
+# Series segments add impedance. Reading a leaf's `tap * x` directly lets a zero-impedance
+# segment contribute exactly 0.0, with no transient `Inf` for the sum to absorb.
```
luke-kiernan, line 196: As below, might be worth checking that these are specializing on type (and if not, add a type parameter). Perf reasons

luke-kiernan: I checked and it compiles just fine as-is

luke-kiernan (changes requested):
One big-picture thing: instead of passing the whole of `nr::NetworkReductionData`, it might be conceptually clearer to just pass the float `_minimum_retained_impedance(nr)`, if that's really all you need in the function body.

Also, is there anywhere you'd want the non-finite series susceptance? I'm wondering if we should just change `get_series_susceptance` to return what-you-currently-have-as `_finite_series_susceptance`.

src/common.jl
m-bossart: This is intentional, it should include the tap

src/common.jl
m-bossart: The issue here is the Ybus path only substitutes when `r==0` and `x==0` whereas the DC path substitutes when `x==0` (regardless of `r` because `r` is already ignored in the DC assumptions). So you're right, this PR is about keeping these things consistent, but its not clear to me a better solution

src/definitions.jl
m-bossart: This is fixed so the public accessors (`get_series_susceptance` `get_series_phase_shift`) only take ACTransmission now and get rid of this union.

luke-kiernan (approved):
Good enough.

Question on big picture design stuff: why did you remove `SeriesSegment` and go to straight `ACTransmission`?

PR #367, opened 2026-09-12
fix: resolve transitive protected-bus merging in zero-impedance reduction

mcllerena:
When running the AC Power Flow on EI cases (2026/2027/2030/2035) for `SLL` and `SUM` scenarios. The following error triggers for AC-VOLTAGE VSC control

> ERROR: Two VSC converters both use AC-voltage control at bus 181730; duplicate |V_ac| pins make the Jacobian singular (or the setpoints infeasible). Use reactive-power control for one of them.
Stacktrace:
  [1] error(s::String)
    @ Base ./error.jl:44
  [2] _validate_vsc_ac_controls(dcn::PowerFlows.DCNetwork, bus_types::SubArray{ACBusTypes, 1, Matrix{…}, Tuple{…}, true})
    @ PowerFlows ~/Documents/marck/GDO/PowerFlows.jl/src/vsc_utils.jl:142
  [3] initialize_DCNetwork!(data::PowerFlows.PowerFlowData{…}, sys::System, bus_lookup::Dict{…}, reverse_bus_search_map::Dict{…}, removed_buses::Set{…}).....
.....

The zero-impedance reduction in PNM.jl was merging two distinct VSC terminal buses through an intermediate bus. PNM already supported irreducible buses, but the protection logic was checked against the original branch endpoints only. After an intermediate bus had already been merged, it was no longer recognized as belonging to a protected group and got reduced.

This PR ensures the VSC endpoint buses remain distinct through the zero-impedance reduction with the VSC control modes preserved so that PF.jl can map the converters to distinct reduced AC buses.


src/apply_zero_impedance_reduction.jl
```diff
@@ -69,6 +69,16 @@ function _is_zero_impedance_arc(
     )
 end
 
+function _find_reduction_root(bus_number, reverse_bus_search_map)
```
luke-kiernan: `_update_bus_maps!` keeps the map flat, so `get(reverse_bus_search_map, bus_number, bus_number)` should be fine.

src/apply_zero_impedance_reduction.jl
```diff
@@ -91,14 +101,20 @@ function get_reduction(
...
-            if from_irred && to_irred
-                @warn "Zero-impedance branch between two irreducible buses $from_no and $to_no; skipping merge."
+            from_root = _find_reduction_root(from_no, nr.reverse_bus_search_map)
+            to_root = _find_reduction_root(to_no, nr.reverse_bus_search_map)
+            from_irred = from_root ∈ user_irreducible
+            to_irred = to_root ∈ user_irreducible
+            if from_root == to_root
+                continue
+            elseif from_irred && to_irred
+                @warn "Zero-impedance branch between two irreducible bus groups $from_no and $to_no; skipping merge."
```
luke-kiernan, line 103: Is our code written such that it handles non-merged ZILs sensibly? These are the sort of edge cases that can cause bugs...the obvious and worst one would be fallout from division-by-zero, but I'm pretty sure that we replace impedances that are zero with some minimum value.

mcllerena: Yes, the division-by-zero is being avoided in this case. For branches with `r = 0` and `x = 0`, `Ybus` substitutes `min_x_eps` before computing the admittance, and `ZIBR` uses the same value when classifying the branch.

src/apply_zero_impedance_reduction.jl
mcllerena: You're right. I've already updated it on the last commit.

src/apply_zero_impedance_reduction.jl
```diff
@@ -91,14 +91,20 @@ function get_reduction(
...
             from_no, to_no = arc_key
-            from_irred = from_no ∈ user_irreducible
-            to_irred = to_no ∈ user_irreducible
-            if from_irred && to_irred
-                @warn "Zero-impedance branch between two irreducible buses $from_no and $to_no; skipping merge."
+            from_root = get(nr.reverse_bus_search_map, from_no, from_no)
+            to_root = get(nr.reverse_bus_search_map, to_no, to_no)
+            from_irred = from_root ∈ user_irreducible
+            to_irred = to_root ∈ user_irreducible
+            if from_root == to_root
```
m-bossart, line 98:
```suggestion
            if from_root == to_root
            # Both ends already in one merged group: the arc is a self-loop, not a skipped merge.
             push!(nr.removed_arcs, arc_key)
```

m-bossart (commented):
This is fixed in the psy6 line in #362. The handling of the zero impedance loops is in #368. We should port the tests included in those 2 PRs here. Other than that, just the one small fix. This PR and #368 should make the handling equivalent in main and psy6

luke-kiernan (approved):
I'll approve but I'd suggest adding a test case, or at least opening an issue to add one

m-bossart (changes requested):
@mcllerena can you port the tests from #362 also

jd-lara (commented):
This PR should be green in CI/CD to pass

PR #373, opened 2026-09-22
Jd/psy6 correctness

jd-lara:
This PR has many fixes from a comprehensive review of psy6 branch and performance improvements


src/apply_zero_impedance_reduction.jl
```diff
@@ -7,8 +7,9 @@ _is_transformer(::PSY.TwoWindingTransformer) = true
 # transformer arcs and must be excluded from zero-impedance bus merging like any transformer.
 _is_transformer(::ThreeWindingTransformerCircuit) = true
 _is_transformer(::PSY.ACTransmission) = false
-_any_transformer(parallel_br::AbstractBranchesParallel) =
-    any(_is_transformer(br) for br in parallel_br)
+# Aggregates subtype `PSY.ACTransmission`, so the blanket above would answer `false` for a
+# group or chain holding a transformer. Recursive, so a nested aggregate resolves too.
+_is_transformer(seg::AbstractReductionAggregate) = any(_is_transformer, seg)
```
luke-kiernan, line 11: Behavioral change: now series will report `true` too. Check effects of that

src/BranchAdmittance.jl
```diff
@@ -425,23 +423,35 @@ function equivalent_branch(
...
+π-model admittance `(g, b, g_fr, b_fr, g_to, b_to, tap, shift)` of a single branch as the
+assembled matrices carry it, where `g + im*b == 1 / (r + im*x)` is the series admittance.
+The admittance-form view of [`equivalent_branch`](@ref); see it for the shunt and unit
+conventions, and for what `nr` contributes.
 
-π-model admittance for a reduction-aggregated arc (a `BranchesSeries` chain or
-`BranchesParallel` group), built from PNM's reduction-aware equivalent physical branch
+There is no `nr`-less form: an admittance that skipped `nr`'s impedance correction did not
+match what was stamped, and mutating the Ybus with it was a live bug rather than a
+hypothetical one.
```
luke-kiernan: Ask Claude to go through the comments and cut what won't be relevant in a month or two.

luke-kiernan (commented):
Unable to untangle what's going on here in reasonable time. Maybe Matt will have a better read. Sources of confusion:

1. Nesting. What's recursive and what's not recursive?
2. Too many different sources of truth: we have helper functions for equivalent impedance/susceptance/admittance, calculated from components vs matrices, edge cases of finite-ness, reversed, DC...
3. Invariants of the network reduction maps. E.g. which can have reversed arcs? Which things are recursed into how (`any` vs `all`)? etc.


I asked AI about some of this. Useful stuff:

> ### Five scalar susceptances
> 
> | # | Accessor | Reads | `r == x == 0` | ICT | Used by |
> |---|---|---|---|---|---|
> | 1 | `get_series_susceptance(b, units)` | stored `1/x`, tap-divided for 2W circuits | **throws** | no | public API; PSY's deleted method |
> | 2 | `_series_susceptance_raw(b, units)` | same | returns `Inf` | no | the single home for the arithmetic |
> | 3 | `_finite_series_susceptance(seg, min_x_eps\|nr)` ≡ `get_effective_series_susceptance` | #2, substituting via `_zero_impedance_susceptance` | substitutes | no | shares, weights, chain arithmetic |
> | 4 | `_ba_arc_susceptance(seg, nr)` | the entry's **own two-port** | inherits Ybus's | **yes** (symmetric arcs) | `BA_Matrix`, outage full/partial test |
> | 5 | `_dc_entry_susceptance(br, min_x_eps)` | = #3, but errors on `GenericArcImpedance` | substitutes | no | `arc_dc_shift_injection` only |
> 
> ### Arc resolution — four probe orders
> 
> Four functions resolve "what occupies this arc", and no two agree:
> 
> | Function | Maps probed, in order | Reverse orientation? |
> |---|---|---|
> | `_resolve_arc_entry` (`common.jl`) | direct → series → parallel → added | series & parallel only |
> | `_reduced_arc_equivalent_branch` (`BranchAdmittance.jl`) | series → parallel | yes |
> | `_compute_arc_ybus_delta` (`network_modification.jl`) | direct → parallel → series | no |
> | `_existing_arc_key` / `_get_branch_map_entries` (`degree_two_reduction.jl`) | direct → parallel (→ series for `_existing_arc_key`) | yes, as separate entries |
> | `get_arc_axis` (`NetworkReductionData.jl`) | direct, parallel, series, added (union) | n/a |
> | `_validate_catalog_closure` | direct, parallel, series (**not** added) | n/a |
> 

> `_segment_susceptance_after_outage` and `_segment_phase_shift_after_outage` test `tripped_set` membership one level deep while the values they sum recurse fully. Tripping a leaf below a parallel-in-series segment returns `Δb ≈ 0` silently, where every other unsupported-nesting path in `network_modification.jl` throws.
> 
> `_get_partial_chain_recursive!` is unbounded-depth recursion on the bus graph, in the same area where `_surviving_root!` was deliberately de-recursed for the 20k-path case.

src/BranchAdmittance.jl
```diff
@@ -50,7 +50,7 @@ end
 # Kept out of line so the error body does not count against the caller's inlining budget.
 @noinline function _throw_non_finite_susceptance(segment::PSY.ACTransmission, b::Float64)
     error(
-        "Series susceptance of $(get_name(segment)) is $(b): the branch has r == x == 0. " *
```
luke-kiernan: Why this edit?

src/BranchCatalog.jl
```diff
@@ -410,10 +406,16 @@ Component-name index for name-based matrix indexing (`get_branch_multiplier`), w
...
-name-based matrix indexing does not resolve them.
+aggregate's entry name is the group's, not any component's.
+
+A name is indexed only where `_branch_multiplier` can answer: the arc's entry or one of its
+direct members. Chain members, standalone or grouped, are never indexed.
 """
-function _build_component_name_index(nrd::NetworkReductionData, predicate)
+function _build_component_name_index(
+    nrd::NetworkReductionData,
+    arcs::ARC_TABLE,
```
luke-kiernan: Why do you need the `arcs` here?

src/BranchesSeries.jl
```diff
@@ -1,18 +1,45 @@
 mutable struct BranchesSeries <: AbstractReductionAggregate
-    branches::Dict{DataType, Vector{<:PSY.ACTransmission}}
-    needs_insertion_order::Bool
+    branches::Dict{DataType, Vector{PSY.ACTransmission}}
     insertion_order::Vector{Tuple{DataType, Int}}
     segment_orientations::Vector{Symbol}
```
luke-kiernan: 3 separate lists, each with one item per branch. If we end up with a 4th, I'd go to one vector of tuples

src/BranchesSeries.jl
```diff
@@ -25,90 +52,24 @@ function add_branch!(
     branch::T,
     orientation,
 ) where {T <: PSY.ACTransmission}
-    # Clear the cached two-port up front so every early return below is covered.
     invalidate_equivalent_ybus!(bs)
```
luke-kiernan: where is the equivalent ybus re-calculated?

src/common.jl
```diff
@@ -674,11 +760,13 @@ function _dc_series_impedance(tw::ThreeWindingTransformerCircuit)
     return _dc_series_impedance(tw.circuit)
 end
 
-function _dc_series_impedance(bp::AbstractBranchesParallel)
+# A member can itself be an aggregate, so inference recurses through these two and gives up;
+# the annotations cut that off at the only place the value is known to be a scalar impedance.
```
luke-kiernan: > the annotations cut that off at the only place the value is known to be a scalar impedance.

I don't follow

src/NetworkReductionData.jl
```diff
@@ -47,67 +38,6 @@ function Base.:(==)(a::BranchMapsByType, b::BranchMapsByType)
...
-    b::BranchMapsByType,
-    ::Type{T},
-) where {T <: PSY.ACTransmission}
-    return b.direct_branch_map[T]::Dict{Tuple{Int, Int}, T}
-end
-
-# `ThreeWindingTransformerCircuit` is non-parametric (one concrete parent type), so the
-# per-type bucket is keyed by the parent transformer type and holds the wrapper. `T`
-# (a concrete `PSY.ThreeWindingTransformer` subtype) selects that bucket.
-function get_typed_direct_branch_map(
```
luke-kiernan: wait, did you get rid of these are move them?

src/populate_cache.jl
```diff
@@ -261,9 +250,9 @@ _resolve_monitored_index(vmodf::VirtualMODF, m::Tuple{Int, Int}) =
     _woodbury_factors_from_base(base_full, BA, arc_sus, modifications, n_bus) -> WoodburyFactors
 
 Assemble Woodbury factors from precomputed pre-contingency solves. `base_full`
-maps each modified arc index to `B⁻¹ · BA[:, arc]` scattered to full-bus space.
-Identical math to `_compute_woodbury_factors_impl` but with the per-arc libklu
-solves replaced by dictionary lookups (the batched solve already paid for them).
+maps each modified arc index to `B⁻¹ · BA[:, arc]` scattered to full-bus space,
+so the per-arc libklu solves of `_compute_woodbury_factors_impl` become
+dictionary lookups; the shared kernel does the rest.
```
luke-kiernan: I don't follow this. Maybe an EE person would, though

src/network_modification.jl
```diff
@@ -1,9 +1,9 @@
-# `delta_b` removes the arc's entire series susceptance (a full outage).
-# `delta_b` comes through the `ComplexF32` Ybus, so the two sides agree only to Float32
-# precision; `sqrt(eps(Float32))` sits well above that noise and well below any real partial
-# outage ratio (an identical double circuit is 0.5).
+# Compared as magnitudes: delta_b from _get_arc_susceptances is -|b|, but _ba_arc_susceptance
+# is signed, so a signed test calls a full outage of a negative-reactance arc (3W star leg,
+# series compensation) partial. Unresolved whether -|b| is right there: MODF disagrees with
+# a direct oracle on such arcs.
```
luke-kiernan: > # Unresolved whether -|b| is right there: MODF disagrees with a direct oracle on such arcs.

umm what does that mean...

src/virtual_factor_core.jl
```diff
@@ -111,21 +121,34 @@ Return the raw diagonal `H[e, e]` of `PTDF · A`, computing it (one solve per
...
 """
+# Double-checked publish for a lazy shared vector: `compute` runs once under `lock` and
+# `dest` is filled in place before `ready` is set, so a throw mid-compute leaves the flag
+# unset and a retry does not append a second copy.
+function _publish_once!(
+    compute,
+    dest::Vector,
+    ready::Threads.Atomic{Bool},
+    lock::ReentrantLock,
+)
```
luke-kiernan: I asked Claude about this. It suggested that the double-checking doesn't really earn its keep, that this would work just as well.

> ```julia
> # Fill once under `lock`; `compute` returns before `append!`, so a throw leaves `dest` empty.
> function _publish_once!(compute, dest::Vector, lock::ReentrantLock)
>     @lock lock begin
>         isempty(dest) && append!(dest, compute())
>         return dest
>     end
> end
> ```

flowgates_spec.md
```diff
@@ -210,10 +210,11 @@ existing grouping.
```
m-bossart: Don't think this file should be committed. Same with `auto_ptdf_tolerance_spec.md`

src/network_modification.jl
jd-lara: THis is to be able to handle the outage of a negative impedance

src/virtual_factor_core.jl
jd-lara: The outer `ready[] && return dest` avoids taking a lock on every read after the first computation; dropping it forces every subsequent read through the lock, which matters since this feeds virtual PTDF/LODF row access on a hot path so I don't think we should implement this change

src/apply_zero_impedance_reduction.jl
jd-lara: This was a bug actually aggregates will not work correctly

src/BranchCatalog.jl
jd-lara: resolves whether a parallel member's arc entry is a plain match or a MixedBranchesParallel

src/BranchesSeries.jl
jd-lara: Opening an issue to discuss

src/BranchesSeries.jl
jd-lara: Lazily, in get_equivalent_physical_branch_parameters/populate_equivalent_ybus!

src/NetworkReductionData.jl
jd-lara: the changes @m-bossart made in POM removed the need for this

